PR #60, opened 2020-05-20
Added JSON format

Logic-gate:
Export as JSON as:

```json

{
     "iso_code":{
                 "name":"value"
      }, 
      ...

}

```

As stated in my comment, If `df_to_json` could be written in pure Pandas--without expanding `json` with iso_code as the data name, it would be much cleaner and will adhere to the dev's uniformity.


edomt:
Thanks @Logic-gate! I've done a bit of refactoring, but writing the complete dataset to JSON is a very good idea indeed.

PR #94, opened 2020-07-02
Add annotations.csv

breck7:
Do we already have cases and deaths annotations in this repo? If not this adds it.

I can hook up Grapher to import entity annotations from this. I added a day column so when we deploy the front end for specific day annotations in Grapher then we are already set for it.

@edomt would this be the correct place to put this file?
@HannahRitchie  would this be easier to update than the current status quo? Anything else we can do?

### DIFF
--- a/scripts/input/owid/annotations.csv
+++ b/scripts/input/owid/annotations.csv
@@ -1,0 +1,12 @@
+location,day,cases_annotations,deaths_annotations
+Spain,2020-04-19,methodology change,
+Spain,2020-04-25,methodology change,methodology change
+Ecuador,2020-05-08,methodology change,
+United Kingdom,2020-05-20,methodology change,
+France,2020-06-02,methodology change,
+India,2020-06-17,,earlier deaths added
+Chile,2020-06-18,earlier cases added,
+Italy,2020-06-25,,methodology change
+United States,2020-06-26,,probable/earlier deaths added
+United States,2020-07-01,,probable/earlier deaths added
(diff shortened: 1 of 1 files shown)


edomt:
That's a great idea I think! This way @HannahRitchie can directly edit them without asking you.

HannahRitchie:
Yes, sounds like a good idea to me. Is the key here that annotations are given in a new single column and allocated to specific day/point for that given country?

edomt:
@breck7 Was the idea that rows with a date would be day-specific annotations, and rows with a blank date would be series-wide annotations?

breck7:
@HannahRitchie, yes, you would provide an annotation for a specific entity+date. For now I recombine the date with the annotation and have 1 annotation per series. Later when we support point annotations there will be no changes needed.

For case fatality rate, I combine the cases + deaths columns into a 3rd column.

Exactly @edomt. If there is no specific date attached to an annotation (such as EU), you can leave the date blank and we treat it as a series annotation both now and when we switch to point annotations. It would be a little simpler for us if these 2 `_annotations` columns were directly in the mega file, like the `tests_units` column, but it's not a big deal to merge these 2 at run time.

edomt:
@HannahRitchie and I would prefer to avoid adding annotations directly in the main CSV file (99% of the rows would be empty, and 1% of them would contain very long strings, and we think that might be pretty confusing for some users).

Other than that, this system sounds great to me!

PR #172, opened 2020-11-04
Update new_cases.csv

xrs1969:
3-11-2020 Greece 2166 new cases


edomt:
Hi @xrs1969 
We source our data directly from the European Centre for Disease Prevention and Control, and report it without making alterations to its original cases and death figures—and the ECDC itself reports them exactly as they are given by national sources. I'll reject this PR as this CSV file is only the result of other scripts being run, and your change would be overwritten at the next update.

xrs1969:
> Hi @xrs1969
> We source our data directly from the European Centre for Disease Prevention and Control, and report it without making alterations to its original cases and death figures—and the ECDC itself reports them exactly as they are given by national sources. I'll reject this PR as this CSV file is only the result of other scripts being run, and your change would be overwritten at the next update.

understood !! have a nice day

PR #238, opened 2020-12-29
Update locations.csv

philiprusinov:
The first official data for  Bulgaria were presented from ministry of health. Till now there isn't automated open data, which could be intergrated.


edomt:
Hi @philiprusinov 
Thank you, we've added Bulgaria to our data.
(For future countries/data, see https://github.com/owid/covid-19-data/issues/230)

PR #251, opened 2021-01-05
Automation to update data on Italian vaccinations

bergwald:
This pull request automates the collection of vaccination data for Italy. #250 

Since Selenium is used to render JavaScript and fetch the data, it is necessary to install ChromeDriver on the machine where the script is executed. The latest version can be found at https://chromedriver.chromium.org/. I've successfully tested the script with the latest stable release (87.0.4280.88). Once downloaded, you just need to point `executable_path` to the ChromeDriver executable.


edomt:
Thank you @bergwald, this is immensely useful! (Especially as the dashboard looks nice but can take a while to load.) I've added a few lines to collect the date as well.

PR #255, opened 2021-01-05
Update Hungary.csv

bdaniel88:


edomt:
Thank you @bdaniel88—we'll add this in today's update. See #230 for future contributions.

PR #266, opened 2021-01-07
Automate vax for United Arab Emirates

lucasrodes:
Added incremental automation for the United Arab Emirates. Used this reference: https://covid19.ncema.gov.ae/en.

Similarly to the [automation for Spain](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/automations/incremental/spain.py) it uses the execution date as the date of the data.

**Minor doubt**: Do we use the date as in UTC, country local time, etc.? For this automation, I have used country local time (note the use of `pytz`) but, for instance, in the automation for Spain it seems that you use the date from the place the script is executed (`datetime.date.now()`)


edomt:
Thank you!

PR #267, opened 2021-01-07
Update Czechia.csv

DawyD:
Update based on the number reported by Czech Ministery of Health on 6.1.2021


edomt:
Thanks @DawyD — we've now automated our data collection for Czechia based on https://onemocneni-aktualne.mzcr.cz/covid-19

PR #271, opened 2021-01-07
Update United Kingdom.csv

bdaniel88:


edomt:
Thanks @bdaniel88 - for the UK our collection is automated, so the data will be automatically added when we update today.

PR #272, opened 2021-01-07
Update United Kingdom.csv

joehoyle:
This is pulled from https://api.coronavirus.data.gov.uk/v2/data?areaType=overview&metric=cumPeopleReceivingFirstDose&metric=cumPeopleReceivingSecondDose&format=csv, which is actually in the automation in https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/automations/batch/united_kingdom.py, but that automation doesn't appear to be running.


joehoyle:
Looks like @bdaniel88 opened this exactly at the same time! https://github.com/owid/covid-19-data/pull/271 -- looks like the above automations run fairly perhaps too. Feel free to close out!

PR #276, opened 2021-01-08
Update Portugal.csv

musicaldoctor:
New government data on vaccination numbers for Portugal


edomt:
Thank you! We'll add this in a minute

PR #277, opened 2021-01-08
Update Romania.csv

wannabeuss:


lucasrodes:
Hi @wannabeuss, for most countries there are scripts that automatically update the vaccination data. These scripts are usually run a couple of times a day and automate the process of scrapping the data sources, cleaning, and updating [the data](https://github.com/owid/covid-19-data/tree/master/public/data/vaccinations/country_data). 

For the case of Romania, there is [this script](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/automations/incremental/romania.py), so its data should be updated throughout the day.

Note that not all countries have such scripts. You can check which countries are automated in the file [automation_state.csv](https://github.com/owid/covid-19-data/tree/master/scripts/scripts/vaccinations/automations).

ℹ️ You can get more details in #230 and #250.

PR #278, opened 2021-01-08
TESTING: Add data & automation for Andorra

lucasrodes:
Reference: https://www.govern.ad/covid19/


edomt:
Thank you venturing into the testing data @lucasrodes :)
One thing we'll need to know before adding this is whether the positive cases (as reported by the country, counted by JHU, and reported in our data) include only PCR or PCR+TMA. Because we need those two units to match between the numerator (# of cases) and denominator (# of tests) to calculate an accurate positive rate.

lucasrodes:
@edomt Will dig into it!

lucasrodes:
After making some numbers I came to the conclusion that **the number of positive cases (as reported by the country in the provided dashboard) should be including PCR and TMA**.

Let me detail my checks below, sorry for the verbosity!

### Confirmed cases: Government's dashboard vs JHU data
As of 13 January 2021, on the [Gov's dashboard](https://www.govern.ad/covid19/) it states:

> 443 / 8.682
> Total casos confirmats en la darrera setmana / Total casos confirmats fins la data d’actualització.

Translates as 8682 confirmed cases until today and 443 last 7 days (last data update was 11 January 2021). When compared with JHU data we get an exact match:

```python
>>> import pandas as pd
>>> url = "https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/jhu/new_cases.csv"
>>> df = pd.read_csv(url)
>>> df["Andorra"].sum()
8682.0
```

Conclusion: The number of cases reported in the dashboard and by JHU should be measuring the same metric.

### Positive cases in dashboard come from PCR+TMA?
To check whether the number of reported confirmed cases come from both PCR and TMA tests, I use the gov's given positivity rate in the last 7 days (reference date is 11th of January 2021), which is 14.5% (using PCR and TMA, as stated below).

> Positivitat PCR / TMA
> 14,5%
> És el quocient entre positius i nombre de test efectuats amb resultat dels darrers 7 dies.

The number of tests performed in the last 7 days equals the difference between the total tests performed on last week's update (11th of January 2021) and total tests performed on the previous week's update (4th of January 2021), i.e. `130657-127577=3080` (numbers provided in Andorra.csv, extracted from [gov's dashboard](https://www.govern.ad/covid19/)).
 
Now, if I compute the ratio between the number of confirmed cases in the last 7 days (i.e. 443, as shown in the first quoted text) and the total number of tests performed in the last 7 days:

```python
>>> 443/(130657-127577)
0.14383116883116884
```

Which roughly equals 14.5% stated before. 

### Conclusion

**We know that the ratio (i.e. positivity rate, 14.5%) and the denominator (i.e. the number of performed tests, 3080) measure PCR + TMA, per definition. Therefore, the numerator (which appears to be 443, the number of confirmed cases) must also be measuring PCR + TMA.**

edomt:
Nice @lucasrodes, thank you for explaning all this!

PR #280, opened 2021-01-09
Update China.csv

ira921208:


edomt:
Wow, very nice catch @ira921208! Thank you. We'll add this in today's update (we can't take contributions directly via pull requests, see #230)

PR #288, opened 2021-01-11
Update Russia.csv

bdaniel88:


edomt:
Hmm, unfortunately we'd had issues with statements made by Gintsburg in the past and decided to no longer use them: https://twitter.com/redouad/status/1343544952052133888
We'd really prefer statements made by the minister or a similar public official rather than the Gamaleya Institute, which seems to be focused on production outputs.

PR #300, opened 2021-01-12
Fix source_url column for Austria

paralaxsd:
This fixes a minor copy paste issue in the data.


edomt:
Thank you @paralaxsd! (and thanks again for the file)
I'll fixe that in the input data

PR #316, opened 2021-01-15
Update Portugal.csv

leonardo-oliveir-a:
New Portugal data.


edomt:
Thank you @leonardo-oliveir-a 
For Portugal we're collecting the data from the official dashboard which is still at 70000 right now: https://covid19.min-saude.pt/ponto-de-situacao-atual-em-portugal/

PR #326, opened 2021-01-16
Update denmark.py

mxsc:
Add info about total vaccinations and fully vaccinated people.

The data for Denmark has been updated with fully vaccinated people from today (16.01.2021) [](https://files.ssi.dk/covid19/vaccinationstilslutning/vaccinationstilslutning-16012021-co42)

The table format has changed a bit, therefore the header manipulation.

`people_vaccinated`: is taken from the column that translates to People who started the vaccination
`people_fully_vaccinated`: is taken from the column that translates to People who finished/ are fully vaccinated
`total_vaccinations`: is not a column in the pdf, but from the vaccinations readme, i took it should be the sum of the people_vaccinated,people_fully_vaccinated.

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/denmark.py
+++ b/scripts/scripts/vaccinations/automations/batch/denmark.py
@@ -14,15 +14,22 @@ def main():
 
     # Fetch data
     df = pd.DataFrame()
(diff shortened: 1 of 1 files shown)


edomt:
Thank you very much @mxsc! This is really useful. I've made one small change and I'll merge this now.

PR #338, opened 2021-01-19
Moderna is used in Czech republic too

flekmatik:
https://www.idnes.cz/ostrava/zpravy/moderna-ockovani-kraj-vakcina-koronavirus.A210115_589622_ostrava-zpravy_woj


edomt:
Thanks, I'll update the metadata accordingly.

PR #340, opened 2021-01-19
Update Sweden.csv

cgbystrom:
Latest data from previously used source.


edomt:
Thanks @cgbystrom! This is already in our database, and it'll go live when our data is updated (~8PM CET).

PR #347, opened 2021-01-21
Automate batch vaccination data for Czechia

kokes:
I copied over the German script and adapted it for Czechia


edomt:
Thank you! Very useful :)

PR #350, opened 2021-01-21
Ireland vaccination numbers data from 21st Jan

donovanh:
Latest figures from https://covid19ireland-geohive.hub.arcgis.com


edomt:
Hi @donovanh. We collect data directly from that website already, so that new data point will be part of today's update (in ~1 hour). But thank you for the contribution!

donovanh:
Cool, thank you.

PR #353, opened 2021-01-22
Update Portugal.csv

bmserras:


edomt:
Thanks @bmserras 
This data will be automatically from https://covid19.min-saude.pt/ponto-de-situacao-atual-em-portugal/ during our daily update.

PR #369, opened 2021-01-26
Update US daily_vaccinations with calculation link

ivanstegic:
@edomt I noticed that the link to explaining the 7 day smoothing was missing on the US version of the explanation of the variable so I added it. Hope this helps.


edomt:
Thank you @ivanstegic !

PR #370, opened 2021-01-26
New vaccination data for Czechia

kokes:
There is a new open data release now - we finally have vaccine names, first/second doses, and other goodies.

(The dataset also contains data split by region and age group, but as far as I understand, OWID only works with aggregated data for most countries.)

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/czechia.py
+++ b/scripts/scripts/vaccinations/automations/batch/czechia.py
@@ -2,7 +2,7 @@ import pandas as pd
 
 
 def main():
-     
+
     source_url = "https://onemocneni-aktualne.mzcr.cz/covid-19"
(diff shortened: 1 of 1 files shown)


lucasrodes:
Hi @kokes, 
Thanks for sharing this! This open data source looks really neat!

On another note, I am trying to integrate regional data on a global scale in this separate [project](https://github.com/sociepy/covid19-vaccination-subnational), which follows a very similar approach to this project and is open for integration in the future. Therefore, this new format is extremely useful!

kokes:
@lucasrodes The dataset contains not only region names, but also NUTS identifiers, so I guess it can be easily integrated into your solution. Feel free to fork this script and use it in your project. Ping me in a PR/issue if you have any specific questions. Also, make sure not to remove (m)any of the asserts - it is not unlikely that the source data will get slightly modified.

edomt:
This is perfect! Thank you @kokes :)

PR #375, opened 2021-01-26
Updating vaccinations data for Costa Rica

fede2cr:


edomt:
Thank you

fede2cr:
You closed this PR, didn't merge it, and then you submitted the exact same patch. What's up with that?

fede2cr:
Actually no. You forgot to remove the fbclid fom the URL. So what is up with that as well?

edomt:
Hi @fede2cr 

If you look closely you'll see that it's not the same commit. You added the data point to the file in the `country_data` folder, which is part of our pipeline's _output_, so it would have been overwritten right at the next update. Our input system is more complicated that this I'm afraid, and that's why we can't take direct pull requests. It's all explained in #230.

Thanks for letting me know about the fbclid though, I'll remove that. 👌

PR #377, opened 2021-01-26
Update Serbia.csv

acasrb16:
New data from January 26th
https://www.b92.net/info/vesti/index.php?yyyy=2021&mm=01&dd=26&nav_category=12&nav_id=1801182


edomt:
Thanks!

PR #381, opened 2021-01-26
Update Serbia

covidalert426:
Serbia 26.01.2021. vaccinated 319.504

https://www.aa.com.tr/ba/balkan/vu%C4%8Di%C4%87-u-srbiji-protiv-koronavirusa-vakcinisano-skoro-320000-ljudi/2123908


edomt:
Thank you @covidalert426, we'll add this

PR #416, opened 2021-02-02
Vaccination data for Maldives

covid19owid:
Adding vaccination data for Maldives

Source:
https://presidency.gov.mv/Press/Article/24281


edomt:
Thank you @covid19owid

PR #421, opened 2021-02-03
Adding Alabania vaccination data

covid19owid:
Adding Alabania vaccination data


covid19owid:
Adding Andorra vaccination data
Source:
https://www.govern.ad/covid19_newsletter/en/

PR #439, opened 2021-02-06
Added Nepal Vaccination data

covid19owid:
Added Nepal Vaccination data

Source:
https://thehimalayantimes.com/nepal/158487-persons-receive-first-dose-of-covid-19-vaccine-in-nepal


edomt:
Thanks @covid19owid! I've updated the number for Nepal—but we can't use the one for Pakistan for now, as the source is anonymous in the article, and we only collect official numbers.

PR #442, opened 2021-02-06
Vaccination in Czechia: new dataset + new vaccine

kokes:
With AstraZeneca being distributed today and planned usage starting on Monday, we can expect it to be in the data on Tuesday. At this point, the script will fail as it checks for unique values in vaccine names.

We don't yet know what "custom" naming the Czech authorities will use, so I've only created a placeholder for now, I'll edit this PR when AstraZeneca appears in the data.


kokes:
As noted in the second commit - we're also getting the data in (yet again) a new dataset. This time it's a CSV with _all_ vaccinations, so we need to aggregate from that.

kokes:
@edomt oh, this was still a _draft_ PR, because I'm awaiting the new vaccine name mapping (should appear tomorrow, AstraZeneca vaccination starts today). But the code should run just fine, I'll just submit a new PR tomorrow.

edomt:
@kokes Oh sorry, I wasn't sure and indeed the code ran just fine, so I thought you might have simply forgotten to request the merge :)

PR #444, opened 2021-02-06
 Automated Western Sahara  vaccination data pull

covid19owid:
Automated Western Sahara  vaccination data pull

Source:
https://covid.hespress.com/fr/


edomt:
Thanks @covid19owid, but these vaccinations are already counted in our database under the location `Morocco`. Of course the situation between Morocco and Western Sahara is complicated to handle in our data, but we cannot double-count these vaccinations by re-counting them for Western Sahara.

Issue #446, opened 2021-02-07
VACCINATIONS: a csv file with fields: country, vaccine, start_date, num_doses


dbaroncelli:
It would be very useful to also have a CSV file with the following fields:
- country
- vaccine
- start_date (first day it started public inoculations in that country)
- tot_doses (total doses of such vaccine inoculated in that country)
- people_fully_vaccinated (the number of people fully vaccinated in that country with that vaccine)

Even if the fields "tot_doses" and "people_fully_vaccinated" is not available for all countries, a record with the other 3 fields will already be very useful.

I guess it would also be useful a csv file with simply the list of all vaccines, stating:
- vaccine name
- vaccine producer
- number of doses required for fully vaccinating a person
- total doses shipped
- total doses inoculated
- people fully vaccinated
- list of countries that started public inoculations of that vaccine

edomt:
Hi @dbaroncelli 

Thank you for the suggestion.
We've now made available a file on doses by manufacturer here: https://github.com/owid/covid-19-data/blob/master/public/data/vaccinations/vaccinations-by-manufacturer.csv
Unfortunately we're far from having access to all the information that you've listed, but that's a start!

PR #463, opened 2021-02-09
Break down by vaccine

edomt:
The idea here is to start working on a system that would allow us to use country data in which different vaccines are separated, and let the system merge them together. This would simplify some of our import scripts, but also provide the necessary data for the kind of file suggested in https://github.com/owid/covid-19-data/issues/446.

The system could now accept a file such as this one:

| location   | vaccine         | total_vaccinations | people_vaccinated | people_fully_vaccinated |
|------------|-----------------|--------------------|-------------------|-------------------------|
| 2021-01-01 | Pfizer/BioNTech | 123                | 123               | 0                       |
| 2021-01-08 | Pfizer/BioNTech | 456                | 456               | 0                       |
| 2021-01-15 | Pfizer/BioNTech | 789                | 789               | 0                       |
| 2021-01-15 | Moderna         | 123                | 123               | 0                       |

And automatically spit out this:

| location   | vaccine                  | total_vaccinations | people_vaccinated | people_fully_vaccinated |
|------------|--------------------------|--------------------|-------------------|-------------------------|
| 2021-01-01 | Pfizer/BioNTech          | 123                | 123               | 0                       |
| 2021-01-08 | Pfizer/BioNTech          | 456                | 456               | 0                       |
| 2021-01-15 | Moderna, Pfizer/BioNTech | 912                | 912               | 0                       |

This solution is _very_ experimental at the moment—I've tested it on the data for Czechia, and it seems to work, but I'd feel much better with 1 or 2 more live examples.

The whole thing would only work when the data is _fully_ disaggregated, i.e. when a country publishes data broken down by vaccine for both 1st dose and 2nd dose data, and I think Czechia is the only country where that's the case. Germany publishes cumulative doses by vaccine [in its file](https://impfdashboard.de/static/data/germany_vaccinations_timeseries_v2.tsv), but without a 1st/2nd dose distinction, which means we can't get `people_vaccinated` and `people_fully_vaccinated` by vaccine.

cc @kokes

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/czechia.py
+++ b/scripts/scripts/vaccinations/automations/batch/czechia.py
@@ -22,26 +22,25 @@ def main():
     # of `total_vaccinations`
     assert set(df["vakcina"].unique()) == set(vaccine_mapping.keys())
     df = df.replace(vaccine_mapping)
+    df = (
+        df.groupby(["datum", "vakcina", "poradi_davky"])
+        .size().unstack().reset_index()
+        .rename(columns={
+            "datum": "date",
+            "vakcina": "vaccine",
+            1: "people_vaccinated",
+            2: "people_fully_vaccinated",
+        })
+    )
 
-    df = df.groupby(["datum", "vakcina", "poradi_davky"]).size().unstack().reset_index()
-    df = df.groupby("datum").agg(
-        vaccine=("vakcina", lambda x: ", ".join(sorted(set(x)))),
-        people_vaccinated=(1, "sum"),  # 1 means 1st dose
-        people_fully_vaccinated=(2, "sum"),
-    ).reset_index()
+    df["total_vaccinations"] = df["people_vaccinated"].fillna(0) + df["people_fully_vaccinated"].fillna(0)
 
-    # the following holds only because all vaccines used so far require two doses
-    df["total_vaccinations"] = df["people_vaccinated"] + df["people_fully_vaccinated"]
-
-    df = df.rename(columns={
-         "datum": "date",
-    })
     df["date"] = df["date"].astype(str).str.slice(0, 10)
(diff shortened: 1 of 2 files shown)


edomt:
Another reason to wait a bit longer is to see how many countries will actually publish that kind of disaggregated data. If it's limited to a handful of countries, then I'd rather not make the system much more complex for a very limited benefit.

edomt:
(Writing down my thoughts)

Another completely different solution, maybe less elegant but easier and more flexible, would be to do the opposite: export data broken down by vaccine inside each country script when it's available. So `czechia.py` would output the normal file in `output` but also another one in an `output_by_vaccine` folder, where we'd have one file for each country that makes this available. 

This means we wouldn't have to deal with all sorts of merging issues when adding up numbers, as the two systems would remain isolated. This would also let us create a first version of this with only doses broken down by vaccine (so including Germany).

kokes:
Having thought about it a little bit:

I think the `output_by_vaccine` solution might be a bit cleaner: it will allow one to easily distinguish between the two datasets - a daily format and the day-vaccine format. They won't have to do any detections of non-uniqueness and other potentially error prone tasks when working with the intermediate dataset (e.g. I oftentimes take the country-level dataset to look at a time series [because it's smaller than the big one], I don't want to do any aggregations).

The code and pipelines will also be sort of separate, so if you decide to backtrack due to lack of data, it will be easier to rollback. Easier than if you had this day-vaccine format in your main pipeline. And since I don't think it's too probable we'll get this level of detail for a meaningful fraction of the countries included, having it separated out makes a little more sense to me.

kokes:
I was told via twitter than Italy seems to have the data necessary for this endeavour. https://github.com/italia/covid19-opendata-vaccini/blob/master/dati/somministrazioni-vaccini-latest.csv

PR #465, opened 2021-02-09
Refactoring proposal

ValentinMouret:
Hey :)

First, thank you very much for your work with Our World in Data. It rocked from day 1 and helped me a lot to understand the pandemic when it started out.

This pull-request is a proposal to streamline the pipelines.
It leverages higher-level Pandas APIs that allow decomposing said pipelines in series of functional steps.

Since steps are decoupled, they are easier to reason about, test, and reuse (e.g. in notebooks for explorations or development or across pipelines). Also, it limits mutable states to a minimum.

Probably it also makes it easier to scan through files to find patterns and set and enforce behaviors across pipelines.

Also, it adds Python type hints, which make it easier to understand the code and leverage autocompletion and such.

If you think there is value in doing that, I can take care of all the pipelines.
Since it takes some time to do, I wanted your feedback on a proof of concept before refactoring all of them. 😄

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/austria.py
+++ b/scripts/scripts/vaccinations/automations/batch/austria.py
@@ -1,31 +1,68 @@
-import datetime
 import pandas as pd
 
 
+def read(source: str) -> pd.DataFrame:
+    return pd.read_csv(source, sep=";")
+
+
+def check_columns(input: pd.DataFrame, expected=14) -> pd.DataFrame:
+    n_columns = input.shape[1]
+    if n_columns != expected:
+        raise ValueError(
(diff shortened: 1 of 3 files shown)


edomt:
Salut Valentin !

This is absolutely great, thank you so much for doing this! Some of this formatting is fairly new to me, so I can't promise I'll be able to 100% match the cleanliness of this in the future scripts I'll write, but I'll try my best ;) 

I'll merge this now (also because I need to make some changes to the script for Austria, and I want to do them on the basis of your improvements), and yes feel free to open another PR (or multiple) for other scripts.

Again, thank you!

PR #471, opened 2021-02-10
Update locations.csv

paolocoda62:
Update on Peru’s vaccination campaign based on Ministry of Health’s data


edomt:
Thank you @paolocoda62! We've now created an automation to fetch the number from that URL.

PR #473, opened 2021-02-11
Refactoring the pipelines for Jersey, Switzerland, Argentina, Faeroe Islands & Liechtenstein

covid19owid:
Refactoring the pipeline based on PR#465

@ValentinMouret  and @edomt 

Can you please review the code.  So that it can be edited/merged with the main branch.


scripts/scripts/vaccinations/automations/incremental/argentina.py
```diff
@@ -1,30 +1,58 @@
...
-    people_fully_vaccinated = int(df["segunda_dosis_cantidad"].sum())
-    total_vaccinations = people_vaccinated + people_fully_vaccinated
 
-    date = str(datetime.datetime.now(pytz.timezone("America/Argentina/Buenos_Aires")).date())
+def add_totals(input: pd.DataFrame) -> pd.DataFrame:
+    return input.assign(
+        people_vaccinated=int(input["primera_dosis_cantidad"].sum()),
+        people_fully_vaccinated=int(input["segunda_dosis_cantidad"].sum()),
+    ).assign(
+        total_vaccinations=lambda df: df.people_vaccinated + df.people_fully_vaccinated,
```
ValentinMouret, line 18: This function is now in the utils. You can now do:

```python
from utils.pipeline import enrich_total_vaccinations

...

.pipe(enrich_total_vaccinations)
```

scripts/scripts/vaccinations/automations/incremental/argentina.py
```diff
@@ -1,30 +1,58 @@
...
-    people_vaccinated = int(df["primera_dosis_cantidad"].sum())
-    people_fully_vaccinated = int(df["segunda_dosis_cantidad"].sum())
-    total_vaccinations = people_vaccinated + people_fully_vaccinated
 
-    date = str(datetime.datetime.now(pytz.timezone("America/Argentina/Buenos_Aires")).date())
+def add_totals(input: pd.DataFrame) -> pd.DataFrame:
+    return input.assign(
+        people_vaccinated=int(input["primera_dosis_cantidad"].sum()),
+        people_fully_vaccinated=int(input["segunda_dosis_cantidad"].sum()),
+    ).assign(
```
ValentinMouret, line 17: This would yield a data frame of size `n`, where `n` is the size of the input, where the columns we add have only one value across rows.
Then later we only take the first item of this data frame.

I suggest:

```suggestion
    return (
        input.sum()
        .rename(
            columns={
                "primera_dosis_cantidad": "people_vaccinated",
                "segunda_dosis_cantidad": "people_fully_vaccinated",
                }
            )
        .pipe(enrich_total_vaccinations)
    )
```

This way, it’s evident we care about the resulting series and not the data frame.

scripts/scripts/vaccinations/automations/incremental/faeroe_islands.py
```diff
@@ -1,31 +1,59 @@
...
 import re
 import requests
 import dateparser
 import vaxutils
+import pandas as pd
 
 
-def main():
+def read(source: str) -> pd.DataFrame:
+    return pd.read_json(requests.get(source).content)
```
ValentinMouret, line 9: This should work:

```suggestion
    return pd.read_json(source)
```

ValentinMouret: Then you could even remove the dependency on `requests`.

ValentinMouret: The JSON we pull is like this:

```json
	{
	"stats": [
		{
			"last_updated" : "11. februar kl. 11:21",
			"cases" : "657",
			"active_cases" : "4", 
			"recovered" : "652",
			"deaths" : "1",
			"quarantined" : "19",
			"tested" : "225.960",
			"awaiting_results" : "1",
			"out_of_quarantine" : "",
			"hospitalized" : "1",
				"first_vaccine_number" : "4134",
	"first_vaccine_procentage" : "7.8",
	"second_vaccine_number" : "1221",
	"second_vaccine_procentage" : "2.3",
	"vaccine_last_update" : "10. februar kl. 09:59"
			}
		]
	}
```

By piping this straight into a data frame, we get a data frame of one row and one column (`stats`).

It would be better to pipe this as early as possible into something that’s easier to consume, e.g.:


```python
    return pd.read_json(requests.get(source).content).pipe(lambda ds: pd.DataFrame.from_records(ds["stats"]).iloc[0])
```

I made it an anonymous function here, it should be another step but the idea is there.
Incidentally, this would make the function return a Series and not a DataFrame.

This would return a series like:

```
last_updated                 11. februar kl. 11:21
cases                                          657
active_cases                                     4
recovered                                      652
deaths                                           1
quarantined                                     19
tested                                     225.960
awaiting_results                                 1
out_of_quarantine
hospitalized                                     1
first_vaccine_number                          4134
first_vaccine_procentage                       7.8
second_vaccine_number                         1221
second_vaccine_procentage                      2.3
vaccine_last_update          10. februar kl. 09:59
Name: 0, dtype: object
```

ValentinMouret: This would make downstream manipulations more straightforward rather that accessing `["stats"][0]` all the time.

scripts/scripts/vaccinations/automations/incremental/faeroe_islands.py
```diff
@@ -1,31 +1,59 @@
...
+            .pipe(format_date)
+            .pipe(enrich_columns)
+    )
+
+
+def main():
+    source = "https://corona.fo/json/stats"
+    data = read(source).pipe(pipeline)
+
+    vaxutils.increment(
```
ValentinMouret, line 49: Similarly, if it becomes a Series, the piece of code below would be simpler.

scripts/scripts/vaccinations/automations/incremental/jersey.py
```diff
@@ -4,32 +4,62 @@
 from bs4 import BeautifulSoup
 import vaxutils
 
+date = None
```
ValentinMouret, line 7: I don’t see why we need a global `date`.
After parsing the table and returning one row off of it, we could enrich the date with `assign`.

ValentinMouret (changes requested):
Thanks for working on this pipeline :)

There is however an overhead because of the difference between batch and incremental pipelines.

It makes sense that batch updates deal with data frames (2D) whereas incremental pipelines deal with Series (1D).

I would suggest that the `read` functions should return a `Series` and that the pipelines consume and return `Series` instead of `DataFrame`.

They also support `pipe` and other methods, so this should yield very similar pipelines.

Please tell me what you think :)

covid19owid:
@ValentinMouret  Thanks a lot !

Really appreciate for the reviews. Will incorporate the changes as you suggested.

covid19owid:
@ValentinMouret  

Thank you for the reviews. Committed files have been updated to use Series and also suggested changes have been made. Can you please let me know your reviews on them.

ValentinMouret (approved):
Great work! Looks good :)

ValentinMouret:
I made some suggestions that could be considered nitpicking.
Super clear now anyways 👍

covid19owid:
@ValentinMouret 

 Thank you very much for the comments and reviews ! . The latest commit has the updated changes made to it.

@edomt , Can you please merge the code at your convenience. Remaining incremental scripts would be updated to Series as per @ValentinMouret suggestions

edomt:
Thank you both so much for your work on this!

PR #478, opened 2021-02-12
czechia vaccination: new vaccine used (AstraZeneca)

kokes:
AstraZeneca was finally used yesterday, so I added a new mapping. Ran the script locally and it worked fine.


edomt:
Thank you!

PR #484, opened 2021-02-13
Make a pipeline for Denmark, Germany, and Greece batch vaccination

ValentinMouret:
## Denmark

Interestingly, it seems that the data were modified.

The pipeline works because the table is indeed the one from the PDF,
it’s just that the data from the PDFs are different between the last
processing and today (even for past dates).

I am not sure about the usual process, so I put the processed data in
the git diff. If needed I can remove it from the commit.

@covid19owid @edomt please review when you have some time :)

Signed-off-by: ValentinMouret <valentin.mou@gmail.com>

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/denmark.py
+++ b/scripts/scripts/vaccinations/automations/batch/denmark.py
@@ -1,60 +1,144 @@
+from typing import Dict
+
 import tabula
 import pandas as pd
+import requests
 from bs4 import BeautifulSoup
-import urllib.request
+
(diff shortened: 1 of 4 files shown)


covid19owid:
Salute @ValentinMouret. Refactoring is excellent !

edomt:
Thanks, I'll merge now.

@ValentinMouret Indeed, there are several countries for which the time series changes on many days at each update, because the date is assigned to the date of vaccination (as opposed to the date of report).

PR #506, opened 2021-02-18
Update Switzerland.csv

miluetolf:


edomt:
Hi @miluetolf 

We report Switzerland and Liechtenstein separately, so the current number (541,031 doses) is already up to date.

PR #509, opened 2021-02-18
Update lithuania.R

camappel:
Update source and add PR (number of positive tests **does not** equal new cases, presumably due to multiple tests per person)


edomt:
Thanks Cam!

PR #532, opened 2021-02-21
Update Maldives.csv

waddey:
Daily updates.


waddey (commented):
The previous tweet was deleted and the commit updated accordingly.

edomt:
Thank you @waddey we'll add this.

Issue #500, opened 2021-02-17
JSON Vaccinations file


RiccardoBorchi:
Hello,

is it possible to create a JSON file for vaccinations?

Thanks.

Best Regards.

edomt:
Hi @RiccardoBorchi 

Do you mean an equivalent of this file? https://github.com/owid/covid-19-data/blob/master/public/data/vaccinations/vaccinations.csv
Or are you referring to another file?

RiccardoBorchi:
Hi @edomt,

yes, I mean a JSON version of that file (vaccinations.csv).

Thank you!

lucasrodes:
@edomt I think this could be really useful for people accessing in an api-like way.

I have recently worked on a similar thing (see https://sociepy.org/covid19-vaccination-subnational/data/api/v1/), where I had vaccination data as CSV and transformed it to get API-like data. 

If you find this useful, I can work on this. I would only need some hints on where to place the script.

A possible format could be:

```json
[
{
  "country": "Albania",
  "iso_code": "ALB",
  "data": [
    {
      "date": "2021-01-12",
      "total_vaccinations": 128,
      "people_vaccinated": 128,
      ...
    },
    ...
  ]
},
{
  "country": "Andorra",
  "iso_code": "AND",
  ...
},
]
```

edomt:
@lucasrodes Can you write in R as well? 😬
Ideally it would be part of [`generate_vaccinations_file.R`](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/generate_dataset.R), probably inside `generate_vaccinations_file()` after line 199.

Don't worry if you can't though! I can probably find time to do that (but in the next week or so, rather than in the next few days).

lucasrodes:
@edomt I will look into this, no problem! 😄 

Long time no R, but happy to get my hands dirty with R again! 

Just let me know if you take this and start working on it, to avoid duplicating efforts

ValentinMouret:
```python
```

ValentinMouret:
How about (for the Python part, I unfortunately don’t know R enough to be useful):

```python
import json

import pandas as pd


def to_json(data: pd.DataFrame) -> list:
    location_iso_codes = {tuple(x) for x in data[["location", "iso_code"]].values.tolist()}
    metrics = [column for column in data.columns if column not in {"location", "iso_code"}]
    return [
        {
            "country": location,
            "iso_code": iso_code,
[...]

PR #534, opened 2021-02-21
JSON: Generate vaccinations.json

lucasrodes:
Proposal for vaccinations.json file generation.

I have added the functionality to generate file `vaccinations.json`, with the format discussed in #500. I finally used R, so the logic was added in [generate_dataset.R](scripts/scripts/vaccinations/generate_dataset.R).

I tested the code separately, but not in conjunction with the other logic as I am missing some files (vax_dataset_config.json). Please check it works as expected.

Let me know what you think, as I do not usually code much with R 😄 !

@edomt @ValentinMouret


ValentinMouret:
Looks like code.
(Don’t know squat about R :D)

edomt:
Thank you so much @lucasrodes.

A few questions for you and @ValentinMouret since you may use APIs and JSON files more often than I do:

- Should the output be prettified or not? File is about 30% bigger with proper indenting.
- I've sorted the result by date - @lucasrodes was there a reason to do it with newer rows first?
- I've excluded NA values from the output instead of outputting them as `""` - is that ok?

lucasrodes:
Hi @edomt,
First of all, thanks for the code improvements... It is extremely helpful to me to refresh/learn R 😄.

On your points:

> Should the output be prettified or not? File is about 30% bigger with proper indenting.
By prettifying it, I think it is easier to read for a human. I use Firefox and have observed that it automatically renders the JSON in a human-readable format, but in Chrome and others, it seems that it is not the case. If we consider that the JSON should be human-readable and the 30% cost in size-increase is assumable, I'd leave it prettified.

> I've sorted the result by date - @lucasrodes was there a reason to do it with newer rows first?
I guess I am used to sort dates from most recent till oldest... But not necessarily better or anything as long as we are consistent

> I've excluded NA values from the output instead of outputting them as "" - is that ok?
I guess is a design decision. I thought of keeping them to ensure all fields are present always. However, this has a cost on memory. I guess it should be OK without them, too

lucasrodes:
@RiccardoBorchi, 
your input would be pretty much appreciated, especially on the JSON format and on the points raised in https://github.com/owid/covid-19-data/pull/534#issuecomment-783369477
Thanks 😄

RiccardoBorchi:
@lucasrodes 

thank you for your work! I am happy to have suggest this update, it is very useful!

About my input, I am not expert, I should use the JSON file in a dashboard on a website that I develop to monitor vaccinations data (https://www.curaitalia.it/vaccinazioni-covid-19). Actually I use CSV version, but with the JSON file it would be much more easier and in this way, if the order of columns change, it would not be a problem (while with CSV file is).

Probably the output, in my opinion, should be prettified and it should be as easy as possible, even for the machine.

I don't know what you mean with "NA" values.

lucasrodes:
Thanks @RiccardoBorchi! Appreciate your input.

> I don't know what you mean with "NA" values.

"NA" refers to "Not Available", which happens when a value is not available for a certain day. 

In CSV format, these appear as empty cells:

|location|iso_code|date      |total_vaccinations|people_vaccinated|people_fully_vaccinated|daily_vaccinations_raw|daily_vaccinations|total_vaccinations_per_hundred|people_vaccinated_per_hundred|people_fully_vaccinated_per_hundred|daily_vaccinations_per_million|
|--------|--------|----------|------------------|-----------------|-----------------------|----------------------|------------------|------------------------------|-----------------------------|-----------------------------------|------------------------------|
|Albania |ALB     |2021-01-10|0                 |0                |                       |                      |                  |0                             |0                            |                                   |                              |
|Albania |ALB     |2021-01-11|                  |                 |                       |                      |64                |                              |                             |                                   |22                            |


However, in JSON, if NA values are excluded, these will not appear in JSON (neither the keys).

```json
[
  {
    "country": "Albania",
    "iso_code": "ALB",
    "data": [
      {
        "date": "2021-01-10",
        "total_vaccinations": "0",
        "people_vaccinated": "0",
        "total_vaccinations_per_hundred": "0",
        "people_vaccinated_per_hundred": "0",
      },
      {
        "date": "2021-01-11",
        "daily_vaccinations": "64",
        "daily_vaccinations_per_million": "22"
      },
     ...
     ]
  },
  ...
]
```

An alternative would be to preserve them with an empty value (which would have a cost in memory):

```json
[
  {
    "country": "Albania",
    "iso_code": "ALB",
    "data": [
      {
        "date": "2021-01-10",
        "total_vaccinations": "0",
        "people_vaccinated": "0",
        "people_fully_vaccinated": "",
        "daily_vaccinations_raw": "",
        "daily_vaccinations": "",
        "total_vaccinations_per_hundred": "0",
        "people_vaccinated_per_hundred": "0",
        "people_fully_vaccinated_per_hundred": "",
        "daily_vaccinations_per_million": ""
      },
      {
        "date": "2021-01-11",
        "total_vaccinations": "",
        "people_vaccinated": "",
        "people_fully_vaccinated": "",
        "daily_vaccinations_raw": "",
        "daily_vaccinations": "64",
        "total_vaccinations_per_hundred": "",
        "people_vaccinated_per_hundred": "",
        "people_fully_vaccinated_per_hundred": "",
        "daily_vaccinations_per_million": "22"
      },
      ...
     ]
  },
  ...
]
```

RiccardoBorchi:
You're welcome, @lucasrodes!

Ok, I understand. So, you can add "NA" values even as empty spaces in JSON, but it is a cost of memory, in fact. If they are excluded, it's okay for me, in the database that value will be NULL.

lucasrodes:
Thanks @RiccardoBorchi.

@edomt, In short, I'd say to keep JSON format prettified and exclude NA values.

RiccardoBorchi:
Hi @lucasrodes,

the world data are missing in the JSON file. Could you insert them?

Thanks!

lucasrodes:
Hi @RiccardoBorchi,
I presume you refer to the entries with `iso_code = "OWID_WRL"`. Could easily add this!

edomt:
@lucasrodes Just did it :) see #563

RiccardoBorchi:
Hi @lucasrodes,

exactly! Now it's ok!

Thank you!

PR #538, opened 2021-02-22
Update vaccination info about DOM

nminaya:
Update vaccination info about Dominican Republic


edomt:
Thanks @nminaya - we can't accept direct pull requests on manually-collected countries (see #230) but we'll add this for the next update.

nminaya:
@edomt Got it! Thanks for updating the data and for pointing me to #230

PR #542, opened 2021-02-23
Automate vaccination data pull for Australia

covid19owid:
Automate vaccination data pull for Australia
File : australia.py


edomt:
Thank you @covid19owid! There's actually an API behind that website, so I've rearranged the script for batch collection instead.

PR #546, opened 2021-02-23
Update costa_rica.R

camappel:
The data isn't available in the explorer. I think this is because the script needs the dplyr and zoo packages

### DIFF
--- a/scripts/scripts/testing/automations/batch/costa_rica.R
+++ b/scripts/scripts/testing/automations/batch/costa_rica.R
@@ -1,3 +1,6 @@
+library(dplyr)
+library(zoo)
+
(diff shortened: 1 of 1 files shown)


camappel:
https://ourworldindata.org/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&country=~CRI&region=World&testsMetric=true&interval=smoothed&smoothing=7&pickerMetric=location&pickerSort=asc

edomt:
@camappel You're right there was a small issue; but it's here:

`fwrite(df, "automated_sheets/Costa_Rica.csv")`

should instead be:

`fwrite(df, "automated_sheets/Costa Rica.csv")`

(without the underscore between "Costa" and "Rica")

PR #552, opened 2021-02-24
Update Cambodia.csv

raingsei:
Updated Cambodian statistics. We might later have to update the last column since many haven't got the second injection yet and the reports by the Cambodian authority do not include that number.


edomt:
Thank you @raingsei 
We'll add this to our input data (we can't merge pull requests directly for manually-inputted countries)

PR #558, opened 2021-02-24
Update locations.csv

LokeshSaini02:
For India: Replaced the data-source to a reliable source on https://www.mohfw.gov.in/ instead of relying on obsolete PIB data.


edomt:
Hi @LokeshSaini02 

We've just made that switch on purpose. https://www.mohfw.gov.in/ does not show data on first and second doses (unless of course we haven't found where).

LokeshSaini02:
Where does that get indicated in the chart though?

edomt:
Do you mean where can you find charts about second doses? If so, you can find them on https://ourworldindata.org/covid-vaccinations (further down the page)

PR #561, opened 2021-02-25
Created file for Ukraine

IhorCo:


edomt:
Thank you @IhorCo 
We can't accept direct pull requests for manually-inputted data (see #230) but we'll add this to our input system.

PR #568, opened 2021-02-26
Automate Ukraine vaccination data pull

covid19owid:
Automate Ukraine vaccination data pull
file : ukraine.py


edomt:
Thank you very much @covid19owid, very useful!

PR #571, opened 2021-02-27
Update New Zealand.csv

jed90:
There have now been over 1000 doses provided as per this article - https://www.tvnz.co.nz/one-news/new-zealand/76k-more-covid-19-vaccine-doses-arrive-in-nz-along-500k-special-needles


edomt:
Thank you @jed90 we'll add this to our manual inputs.

PR #575, opened 2021-02-27
Automate Russia vaccination data pull

covid19owid:
Automate Russia vaccination data pull
 file : russia.py


edomt:
Thank you! :)

PR #579, opened 2021-02-28
Update Panama.csv

mmorenoc25:
Update of the vaccines of March 26 and 27, 2021 with the Pfizer-BioNTech vaccines in the Republic of Panama reported by the Panamanian press in the newspaper "La Estrella de Panamá"


edomt:
Thank you @mmorenoc25 — I'll add this to our input database

PR #584, opened 2021-03-01
Update New Zealand.csv

jed90:
Updated numbers of vaccines now 1500 as per this article - https://www.rnz.co.nz/news/national/437432/covid-19-vaccinations-at-port-of-tauranga-begin


edomt:
Thanks @jed90, we'll add this to our data.

PR #590, opened 2021-03-02
Update Czechia.csv

radekpy:


edomt:
@radekpy Our data collection is automated for Czechia, so this will be automatically added during today's update.

PR #595, opened 2021-03-03
Add latest data from NZ Ministry of Health

NoodlesNZ:


edomt:
Thanks @NoodlesNZ we'll add this to our data.

PR #596, opened 2021-03-03
Update New Zealand.csv

jed90:
9431 people have been vaccinated as per this article https://www.rnz.co.nz/news/national/437540/live-chris-hipkins-and-ashley-bloomfield-give-update-on-covid-19-vaccines


edomt:
Thank you @jed90

PR #597, opened 2021-03-03
Update data for El Salvador & Honduras

covid19owid:
Update data for El Salvador & Honduras


edomt:
Thanks @covid19owid 
I've added these data points to our manual inputs.

PR #620, opened 2021-03-04
Update Dominican Republic.csv

otyfrank:
There is new data about the vaccination process in the Dominican Republic.


edomt:
Thank you! We'll add this.

PR #631, opened 2021-03-05
Uruguay, data source and data updated

3dgiordano:
I propose to change the data source as well as its values to another official source that is now available by the official institution of public health of the country.

Now it is possible to keep the daily record of vaccinated and not just the total.


edomt:
Closing in favor of https://github.com/owid/covid-19-data/pull/632

PR #632, opened 2021-03-05
UY Batch Automation

3dgiordano:
Currently, the Uruguayan Ministry of Public Health has a vaccination status monitoring portal.
The portal is a web page that has an image and that shows the status at the moment as well as the totals of the previous days.
https://www.gub.uy/uruguaysevacuna/Vacunas%20Covid.png

While it is possible to carry out an image analysis job and create an incremental automation process, the quick solution that I propose is to consume a spreadsheet that I keep updated every day with the data provided by the official entity.
It will surely minimize the manual work they have today.

As soon as I achieve an incremental solution or the entity provides a data source that can be scrapped, I will incorporate the method to not depend on my spreadsheet.


edomt:
Thank you @3dgiordano, that's very useful! The main issue with the official dashboard (and the image as well) is that it seems to be restricted to IP addresses in Uruguay — people outside the country (myself included) get a 403 error when trying to access it.

So yes, your external spreadsheet would be of great use until they make the dashboard available from outside the country and/or they publish a CSV file.

One small problem though is that your `people_vaccinated` column seems to be a daily difference between `total_vaccinations`, which isn't what our system expects. `people_vaccinated` should be a count of the number of people who have received at least 1 dose of the vaccine (so given that Uruguay started recently, for now it should be equal to `total_vaccinations`). You can find a longer presentation of our metrics here: https://github.com/owid/covid-19-data/tree/master/public/data/vaccinations#an-example-of-how-we-calculate-our-metrics

3dgiordano:
Hi @edomt 

Ok, now I understand, thanks for sharing the example.

I already fixed the datasource formula to handle the same criteria as the old csv.

I can incorporate the people_vaccinated column right now and I will see how to obtain the fully vaccinated data when we are in those second dose acts.

I updated the csv, and you can run the batch and see if the access to the spreadsheet is correct.

francocorreasosa:
Vaccination data could be pulled out directly from https://msp.gxportal.net/data-vaccine.json

3dgiordano:
@francocorreasosa the use of data-vaccine.json from gxportal.net is discouraged, its use is merely for a graphical perspective of the coronavirus.uy application and is not intended to be the source of information for other reports.

The idea behind the proposed method is to isolate owid people from possible integration problems.
In the future I will change the source data to a data source in a public source code repository. 
In this way there will be more transparency in how the data is obtained, its sources and its review and publication process.

PR #633, opened 2021-03-06
Automate Colombia (Batch)

juanpisuribe13:
I'm back again with another pull request on Colombia. This time, it's automated.

Colombia's Ministry of Health sadly doesn't report the amount of vaccinated people in a .csv file... yet. 
Instead, they're hosting them on the worst way possible... an image. And no, they don't leave it on text in their website which really sucks honestly.

Due to this, this script scrapes the image from their website, uses Tesseract OCR to scan the amount of applied vaccines, and finally adds it to the .csv file.
**And yes, this means that Tesseract (and also Pillow and py-tesseract, which still needs Tesseract anyways) SHOULD be installed.**
When the Ministry of Health finally hosts it on .csv, I'll try to update this script as soon as I can (or if I can't due to school I'll submit a new issue with the .csv URL).
I could also make a .csv file and update it as soon as I can, but due to school I can't keep up with that responsibility.

One little neat feature about this is that if the amount of total vaccinations detected in the image are the same as the last row in the .csv file, it will return an error and stop the script.

I'd recommend to run the script at 7 or 8 PM GMT -5 (Colombian Time), since the Ministry of Health first reports it at 6/7 PM GMT -5 on their Twitter but takes like an hour to update it on their website (which is where the script is scraping for).


userlerueda:
Do we need to include `beautifulsoup4`, `requests`, `pytesseract` and `lxml` in [requirements.txt](https://github.com/owid/covid-19-data/blob/master/scripts/requirements.txt)?

scripts/scripts/vaccinations/automations/batch/colombia.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Eliminates all the involved strings if there's any
+    new_vaccinations = []
+    for character in appliedDoses:
+        try:
+            new_vaccinations.append(int(character))
+        except: # easiest way to detect if it's an integer or not, am i right?
+            continue
+    new_vaccinations = int([''.join(str(num) for num in new_vaccinations)][0])
+    # print("Vaccinated: %i" % (new_vaccinations))
```
userlerueda, line 41: You could also use a regular expression to match digits only:
```suggestion
    new_vaccinations = int(re.sub("[^\d]", "", appliedDoses))
```

But you will need to `import re` library

juanpisuribe13: On that.

juanpisuribe13:
> Do we need to include `beautifulsoup4`, `requests`, `pytesseract` and `lxml` in [requirements.txt](https://github.com/owid/covid-19-data/blob/master/scripts/requirements.txt)?

Yep! Pillow too.

juanpisuribe13:
Okay... I was testing it just now since the Colombian Government released new data few minutes ago and whoops.. Tesseract doesn't want to pick up the numbers for some reason.
image

Weirdest thing is that it was recognizing the numbers with yesterday's data:
image
Let's be logical for a second though, if it didn't recognize yesterday's data I wouldn't make this pull request at all, but oh well.
Here's a comparison between the images of the data of yesterday, and the data of today:
image
No difference at all. Tesseract is so weird.

I'm thinking of using OpenCV to manipulate the image a little bit just so Tesseract can pick it better. If this doesn't work, I guess I'll have to close down this pull request :/. Will update you guys about this.

juanpisuribe13:
I'm back. I'm in a hurry now but to make things short, I made it so the image first gets cropped to the applied doses position, then setting it to gray and do some magic with opencv, so that tesseract can comprehend it much easily.
If you wanna see an example:
image
Of course, if you want to check the processed image by yourself you can uncomment the 32 line in colombia.py

P.S: I've tested it with older data too and it works perfect! hopefully tesseract doesn't confuse it with other numbers in the future

by the way, reminder:
> When the Ministry of Health finally hosts it on .csv, I'll try to update this script as soon as I can (or if I can't due to school I'll submit a new issue with the .csv URL).

juanpisuribe13:
And just so you know, it works perfectly fine with the data from today. So I guess there won't be any problems with this one
image

edomt:
Thank you @juanpisuribe13, this is great! I've just tested it and indeed it works fine for me locally. I've adjusted the script to follow our usual logic for incremental collection, and use our `vaxutils.increment` function.

PR #645, opened 2021-03-07
Fix South Korea automation & update vaccine list

covid19owid:
Fix South Korea automation & update vaccine list


edomt:
Thanks @covid19owid - is something wrong with the current scraper? It seems to be working fine on my end.

covid19owid:
Hi @edomt . Looks like it is working now. Earlier was getting an error that required element or class was not found. Will close this PR.

PR #652, opened 2021-03-08
Update Dominican Republic.csv

otyfrank:


enmanuelrm:
Please also Update the correct estimation of the Dominican Republic population is 10,535,535 https://www.one.gob.do/publicaciones?ID=1463#   this is from official source.

otyfrank:
The population estimates for all countries are based on the United Nations World Population Prospects, and the latest revision was in 2019. The official source you suggest is a great option, but it was published in 2016.

PR #668, opened 2021-03-10
Update New Zealand.csv

jed90:
18,000 people now vaccinated as per this article - https://www.rnz.co.nz/news/national/438064/covid-19-vaccine-2m-new-zealanders-will-get-first-dose-within-four-months

p.s. I've emailed the NZ Ministry of Health to see if we can get a data feed however they are yet to respond.


neon-ninja:
Date should read 2021-03-10, not 2021-03-03

PR #670, opened 2021-03-10
Update Panama.csv

mmorenoc25:
Covid-19 vaccination update in the Republic of Panama corresponding to March 9, 2020 reported in the newspaper La Estrella de Panamá.


edomt:
Thank you @mmorenoc25! Note that we collect this daily infographic from La Estrella each morning, so you won't need to add it every day :)

PR #674, opened 2021-03-10
Update vaccinations-by-manufacturer.csv

jesselau211:


edomt:
@jesselau211 This file is generated automatically - is there a centralized file that makes this data available for HK?

jesselau211:
Sadly, the data only update in separate press release from government only.

PR #692, opened 2021-03-12
Update Panama.csv

mmorenoc25:


edomt:
@mmorenoc25 We collect the data directly from the Estrella infographic, so no need to add it here (unless it's still missing after several days of course!)

mmorenoc25:
> @mmorenoc25 We collect the data directly from the Estrella infographic, so no need to add it here (unless it's still missing after several days of course!)

Thank you so much, i add with the Ministry of Health of Panama because La Estrella today don't have the infographic of the vaccination of March 11, 2021

PR #696, opened 2021-03-12
Update Romania.csv

horea80:
Filled in 11th-12th of March 2021


edomt:
@horea80  Thank you for adding this, however our data collection for Romania is already automated so they'll be added automatically :)

horea80:
> @horea80 Thank you for adding this, however our data collection for Romania is already automated so they'll be added automatically :)
Good to hear @edomt, I was under the impression it was two days behind.

PR #697, opened 2021-03-12
Colombia.py Fix

juanpisuribe13:
I assume colombia.py broke since the government changed the style of the image just so they can fit in the second doses part causing some problems with pytesseract, which is why I'm here.
I fixed that, did some small tweaking, and added support for the second doses.

The government counts both first AND second doses in the total vaccinations part, so for people_vaccinated it subtracts total_vaccinations with people_fully_vaccinated. Feel free to correct it if it isn't that way.

~and no, the government hasn't reported vaccine data separate from an image yet 😔.~
Cheers!


edomt:
Thank you so much @juanpisuribe13!

PR #706, opened 2021-03-13
Fix colombia.py (again)

juanpisuribe13:
The government changed the style of the image again so I changed the position and it works fine now.
I am sorry for making another pull request, I just had to fix this little unexpected error.

P.S: The government started to report data a day later for whatever reason so maybe the script should subtract the day to one (I would do this if I wasn't busy with stuff like school, since doing these big things consume a lot of time for me), buuut at one point they will start reporting it the same day again. It's unpredictable honestly.


edomt:
Thank you (again) @juanpisuribe13 :)

PR #707, opened 2021-03-14
Migrate UY worksheet to data on the GitHub project

3dgiordano:
Hi @edomt 

Now the project behind the template owns a project on GitHub.
It is a repository under construction, but for the purposes of Our World in Data it is already functional.

Calculations and formatting are still done on the spreadsheet.
The difference is that now there are processes that run automatically and are responsible for obtaining the data and updating the spreadsheet (no more from my machine).
Then the csv is generated and automatically versioned in the repository.
Now GitHub versioned the file and provides transparency in updating data. 
I also leave public how we do the data collection of the data from the origin portal.
As the portal changes the data, we update the algorithms and update by hand while we solve the problem.

The use of the spreadsheet also allows manual adjustments and actions that generate the recalculation of values for certain days.
The use of a spreadsheet is still quite versatile.

The worksheet now also has its own dashboard and information about the project.
https://cutt.ly/9zYh13U

Regards.
David

### DIFF
--- a/scripts/scripts/vaccinations/automations/batch/uruguay.py
+++ b/scripts/scripts/vaccinations/automations/batch/uruguay.py
@@ -3,7 +3,7 @@ import pandas as pd
 def main():
 
     df = pd.read_csv(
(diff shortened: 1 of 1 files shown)


edomt:
Thank you David! :)

PR #715, opened 2021-03-15
use legacy domain for now, need to port code to current API

ftrotter:


edomt:
Thank you @ftrotter! We weren't aware of this at all, so this is very useful.

PR #718, opened 2021-03-15
Error in vaccine column (UY)

3dgiordano:
@edomt 

Sorry, the previous commit had an error in the vaccine column.
In the rush of manual correction.
I will be more careful in the future.


edomt:
That's alright, I fixed it on master directly.

PR #720, opened 2021-03-15
Update Colombia.csv

userlerueda:
Update people_fully_vaccinated, we might have a problem with the automated script.

Source: https://twitter.com/MinSaludCol/status/1371249991663153152

EweomP_XAAAohO3


juanpisuribe13:
I think it was this chunk of text that appeared, causing the second doses part position to change.
image
In my opinion, the Colombian government shouldn't report data in an infographic but oh well... [There is hope though](https://www.infobae.com/america/colombia/2021/03/14/le-piden-a-minsalud-crear-un-tablero-con-datos-abiertos-sobre-los-avances-de-vacunacion-covid/)

edomt:
Thank you! Don't worry too much about fixing the OCR - I hope they'll report with an actual file at some point soon, and it's not a huge amount of work for us to do the collection manually until then if needed.

PR #722, opened 2021-03-16
Update Dominican Republic.csv

mmorenoc25:
Update of the vaccination process in the Dominican Republic against covid-19 as of March 15, 2021


edomt:
Thank you

PR #726, opened 2021-03-16
Update Vietnam.csv

Ltn123456:


edomt:
Thank you @Ltn123456 that's very useful - we'll add all these observations to the data today.

PR #732, opened 2021-03-16
Update Panama.csv

mmorenoc25:
Covid-19 vaccination update in the Republic of Panama as of March 15, 2021, by the Ministry of Health of the Republic of Panama. Is published in the Ministry of Health because The newspaper La Estrella de Panamá did not publish the graph on the morning of Tuesday, March 16, 2021.I would also request that all the information be from the Ministry of Health as the official entity in charge of the country's health and not from a newspaper.


edomt:
Thank you, we'll add this.
The reason we've been using the tweets from La Estrella is that:
- They directly quote the figures used by the Ministry of Health
- Their format is much more easy to understand than the MoH tweets, where the number is in a block of text that changes every time.
What would ideally happen is that the MoH would start publishing this somewhere other than Twitter, so that we can collect it automatically.

mmorenoc25:
Thank you very much, the Ministry of Health has available in conjunction with the National Authority for Government Innovation (AIG) the Vacunometer page, however this is outdated for which they have chosen to report on the vaccination in daily communications on twitter and on the official website of the Ministry. In any case, until the vaccination meter is enabled I do not think there is a problem in continuing with La Estrella de Panamá or with the Ministry's Communiqué, in any case I make myself available with everything that may be required. Good morning and Greetings from Panama

PR #734, opened 2021-03-16
Colombia.py position update

juanpisuribe13:
I just changed the position of the Image (and hopefully it will stay that way; never ever report data exclusively on an image kids!), updated the output file and changed the datetime day part from 2 to 1 (feel free to change it back again). That's all.

Also I would recommend running this between 12PM-2 PM since the government is now reporting data from a day before during that time, which is good in my opinion to get accurate information from that specific day. 
It doesn't need to be run at the day I'm writing this, since it's already up to date.

(also if you were wondering what that other pull request was, I was trying to merge some stuff to keep my fork repository up to date, but GitHub messed up.)


edomt:
As usual, thank you!

PR #740, opened 2021-03-17
Update Panama.csv

mmorenoc25:
Update of the vaccination against COVID-19 in the Republic of Panama corresponding to March 16, 2021 in Communiqué # 385 of the Ministry of Health of Panama on its Twitter account.


edomt:
Thank you

PR #746, opened 2021-03-18
Update covid-testing-latest-data-source-details.csv

Ltn123456:
According to WHO, as of 14 March 2021, 2,684,679 tests have been conducted to since the beginning of the outbreak. There seems to be differences between WHO's reports (using the terminology ""tests conducted"") and Ministry of Health's bulletins (using ""samples tested""), possibly due to pooled testing. Ministry of Health also used the terminology "lượt người xét nghiệm" (meaning the number of times people tested), which I still have not been able to figure out what it referred to. For example, [the 17 March 2020 bulletin](https://ncov.moh.gov.vn/web/guest/-/6847912-112) reported that as of 15 March 2021, the country had carried out the RT-PCR test with 2,482,302 samples, equivalent to 3,248,873 times of people tested, of which 2,559 were identified positive. The statistic referring to the times of people tested is higher than the number of samples tested. This means one sample can be tested multiple times? Maybe when a pooled test result is positive or indeterminate, all the specimens in the pool need to be retested individually.


camappel:
Thank you for pointing out this discrepancy. As you can see in our source description, this has been an issue for a while: 

"Some bulletins report the cumulative number of people tested alongside the cumulative number of samples tested. Oddly, the cumulative number of people tested is larger than the cumulative number of samples tested. For example, [the 18 September 2020 bulletin](http://vncdc.gov.vn/vi/phong-chong-dich-benh-viem-phoi-cap-ncov/14061/ban-tin-cap-nhat-dich-benh-covid-19-ngay-18-9-2020) reports 1,183,341 samples tested to date using PCR, while at the same time reporting 1,358,575 people tested to date."

I've looked into it and I believe you are correct that one sample can be tested multiple times. In the article you shared (https://ncov.moh.gov.vn/web/guest/-/6847912-112), it states that "the number of rooms capable of testing for ... SARS-CoV-2 using RT-PCR technique is 157 rooms, the maximum testing capacity is about 62,593 samples / day", while "the number of rooms that have been allowed for confirmed testing is 101 rooms with a maximum testing capacity of about 50,663 samples per day." In another article (https://ncov.moh.gov.vn/web/guest/-/6847912-109), it is stated that "in case of suspected positive, positive must immediately transfer the sample to the laboratory that has been allowed to test confirm SARS-CoV-2".

The percentage of total testing capacity that can be used to confirm tests is about 80%; likewise, the percentage of samples to the number of times people have been tested is about 75% (these ratios have presumable changed over time. As such, it's likely that the excess number of times that people have been tested is due to suspected positive samples being retested in labs that can confirm cases.

However, this would suggest that about 800,000 suspected positive tests were retested in 'official' labs, of which only 2,559 were positive. I therefore agree with your theory that if just one test in a pooled test is positive or indeterminate, all the specimens in the pool need to be retested individually.

I will update our source description to record this finding.

PR #755, opened 2021-03-19
Update Croatia.csv

mjuraga:
Adding new information from Croatia based on national task force press conference from Mar 19th 2021.


edomt:
Thank you @mjuraga, we'll add this

PR #757, opened 2021-03-19
Data for 2021-03-19

mouhib876:
https://www.facebook.com/186480324724413/posts/3958312024207872/
We reached 6861 people vaccinated on 2021-03-19


edomt:
Thank you @mouhib876, we'll add this right away

PR #760, opened 2021-03-20
Update Panama.csv

mmorenoc25:
Covid-19 vaccination update in the Republic of Panama on March 19, 2021 published by the newspaper La Estrella de Panamá


edomt:
@mmorenoc25 Thank you but don't need to collect data for Panama, we already do it on our side.

PR #771, opened 2021-03-21
2021-03-21

mouhib876:
We have reached 12496 people vaccinated on 21-03
This is the link for the update
https://www.facebook.com/186480324724413/posts/3963594130346328/


edomt:
Thanks

PR #776, opened 2021-03-22
21-03-2021

mouhib876:
We are also start using Pfizer/BioNTech vaccine and this is the link 
https://www.facebook.com/186480324724413/posts/3964980620207679/


edomt:
Thank you @mouhib876

PR #779, opened 2021-03-23
Update Vaccination Data for Malaysia

limhenry:
Update Vaccination Data for Malaysia: 21 March and 22 March 2021


edomt:
Thank you

PR #792, opened 2021-03-24
Update Philippines.csv

karljamoralin:
AstraZeneca vaccines arrived in the Philippines last March 4 and has been used since then.

Source: https://www.cnnphilippines.com/news/2021/3/4/AstraZeneca-vaccine-COVAX-facility-arrival.html


edomt:
Thank you @karljamoralin, we'll change that

PR #797, opened 2021-03-24
Update vaccinations.csv

ira921208:
Source:https://www.reuters.com/article/us-health-coronavirus-china-vaccines/china-administered-80-46-million-covid-19-vaccine-doses-as-of-end-march-22-idUSKBN2BG04O and https://news.yahoo.com/china-administered-82-85-million-072740768.html


edomt:
Thank you!

PR #807, opened 2021-03-25
Update Panama.csv

mmorenoc25:
Update of the vaccination against COVID-19 in the Republic of Panama corresponding to March 22, 23 and 24, 2021 reported by the Ministry of Health on its Twitter account. The newspaper La Estrella de Panamá has not published information on its social networks or in the newspaper, so it opted for the information contained in the daily Communiqués of the Health Authority.


edomt:
Thank you

PR #812, opened 2021-03-25
Update China.csv

Cahcn:
Source: National Health Commission of the PRC


edomt:
Thank you @Cahcn, we'll add this today

PR #816, opened 2021-03-26
Update Panama.csv

mmorenoc25:
update of the vaccination against covid-19 in the Republic of Panama to March 25, 2021 updated by the Ministry of Health. The source of the Ministry of Health will be used momentarily.Since the newspaper La Estrella de Panamá has not published the updates since the 20th


edomt:
Thanks

PR #826, opened 2021-03-26
Create internal columnar JSON files for Explorer

danielgavrilov:
Splits up the megafile into multiple files in the `public/data/internal/` folder.

Added notices for others not to rely on this. For now the `public/` folder published on Netlify will keep being the "source of truth" for the COVID Explorer. At some point in the future we may change this to lower bandwidth costs.


danielgavrilov:
After this is merged, we need to:

- Delete the `public/data/owid-covid-data.columnar.json` file & commit (I didn't do this because it would create conflicts)
- Pull the master branch on the server and rerun the megafile script & commit results

PR #841, opened 2021-03-28
Update Guatemala.csv

mmorenoc25:
Covid-19 vaccination update in the Republic of Guatemala as of March 26, 2021 by the Ministry of Public health


edomt:
(Our data collection for Guatemala is already automated.)

PR #843, opened 2021-03-29
Update Panama.csv

mmorenoc25:
Update of the vaccination against COVID-19 in the Republic of Panama corresponding to March 28, 2021 by the Ministry of Health. Government Health and technology authorities what next month they will have the vaccine meter in real time


edomt:
@mmorenoc25 We are already collecting the data for Panama each day on our side, so there is no need to create pull requests for it.

PR #854, opened 2021-03-30
Update Panama.csv

mmorenoc25:
Vaccination Update against COVID-19 in the Republic of Panama corresponding to March 28 and 29


edomt:
@mmorenoc25 Thank you, but we do not need updates for Panama, our team is already doing this each day.

PR #855, opened 2021-03-30
Update Dominican Republic.csv

mmorenoc25:
Vaccination update against COVID-19 in the Dominican Republic corresponding to the week From March 23, published on March 29 by the Ministry of Public Health of the Dominican Republic


edomt:
@mmorenoc25 Thank you, but we do not need updates for the Dominican Republic, our team is already doing this each day.

PR #856, opened 2021-03-30
Update Colombia.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Colombia corresponding to March 28, 2021 reported by the Ministry of Health and Social Welfare


edomt:
@mmorenoc25 Thank you, but we do not need updates for Colombia, our team is already doing this each day.

PR #857, opened 2021-03-30
Update Bolivia.csv

mmorenoc25:
Vaccination update against COVID-19 in the Plurinational State of Bolivia for March 28 and 29 by the Ministry of Health and Sports


edomt:
@mmorenoc25 Thank you, but we do not need updates for Bolivia, our team is already doing this each day.

PR #858, opened 2021-03-30
Update vaccinations.csv

ira921208:
Source: http://www.nhc.gov.cn/xcs/yqjzqk/202103/7289a596c0994109841cb77e2c748a6f.shtml


edomt:
Thanks

PR #863, opened 2021-03-31
Automate vaccination data pull for Bhutan

covid19owid:
Automate vaccination data pull for Bhutan


Technosword (approved):

edomt:
Thank you @covid19owid - unfortunately it looks like this page is no longer updated since March 27, while the official Facebook account has been updated several times. I'll close this for now, but we can come back to it if this page gets updated again.

PR #868, opened 2021-03-31
Dominican Republic.csv Correction

otyfrank:
Hi @edomt I think that the last two tweets haven't been interpreted correctly.  
For 2021-03-30 the tweet says "To date, we have applied a total of 910,869 doses, of which 54,395 correspond to the second administration". Based on that there must be 856,474 people with at least one dose.
I think that the same logic must be used for 2021-03-29 because otherwise we would get a negative change (comparing March 30 with 29) in people with at least one dose.
Or maybe I haven't interpreted your fields correctly. Please let me know if that is the case. Thank you.


edomt:
Hi @otyfrank 

The first one seems fine to me:

> Al día de hoy, hemos vacunado 870,134 personas, de las cuales 20,060 han sido inmunizadas con la segunda dosis... y seguimos vacunando.

So 870,134 people have been vaccinated, and 20,060 are fully vaccinated, for a total of 890,194 doses administered.

For the second one though, you're right:

> Al día de hoy, hemos aplicado un total de 910,869 dosis, de las cuales 54,395 corresponden a la segunda administración ... y seguimos vacunando.

This means that 910869 total doses have been administered, including 54395 second doses — so `people_vaccinated` should be 910869 – 54395 = 856474

We'll fix this now, thank you for letting us know.

otyfrank:
Thank you @edomt 
But by only fixing the second one now we have a negative change in people vaccinated: 870134 and then the next day 856474. People with two dosis increased by ~34k but total dosis only increased by ~20k. That's why I think the first should be interpreted as the second. I think that the first day (of publishing two dosis) the local media misinterpreted the numbers, and that's why they changed the wording for the next day.

edomt:
Yes, you're right!

PR #876, opened 2021-03-31
Tunis

camappel:
Script to retrieve cumulative total of people tested in Tunisia. I will delete the Python batch automation file once this has been merged.


edomt:
Thank you @camappel! I've fixed a couple of small things & merged.
Given that the units seem different from our old series, I've started a new output file with just 1 row - we should make sure that the detailed description Tunisia explains this.

camappel:
Thanks. I'm looking into the units issue at the moment - it's possible that it's the same data but with a different name.

PR #879, opened 2021-04-01
Vaccination/Czechia: AstraZeneca renamed to VAXZEVRIA

kokes:
After the vaccine [was renamed](https://www.ema.europa.eu/en/documents/product-information/vaxzevria-previously-covid-19-vaccine-astrazeneca-epar-product-information_en.pdf), the Czech health ministry renamed it in its data as well.

The pipeline checks unique values in the `vaccine_name` column, so it was crashing, it isn't with this patch.


edomt:
Thank you!

PR #904, opened 2021-04-03
Update Iran.csv

mmehreganfar:
250000 Doses of chinese vaccine received


edomt:
@mmehreganfar Thank you but we don't track doses received, only doses administered.

PR #934, opened 2021-04-06
Update 04-06-2021 Philippines.csv

Jaspac5:
Apologies there was an error in the data. The total vaccinations should've been 855,457 and total people who received the first dose is 826,607 based on the source from the government. (As of an April 6 press briefing)

Source: https://www.facebook.com/ntfcovid19ph/photos/pcb.290001055995828/289998979329369/


edomt:
Thanks, we'll make the change.

PR #950, opened 2021-04-07
Vax/new zealand

lucasrodes:
Added New Zealand automation.


edomt:
@lucasrodes Sorry, I probably didn't give you enough info: there's actually a `Date` tab in the spreadsheet that gives the full time series of vaccinations. We should use this to build a batch collector (instead of an incremental collector based on the `Summary` tab).

PR #968, opened 2021-04-08
Add 2021-04-08 data

santiagob:


lucasrodes:
Uruguay updates are [already automated](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/automations/batch/uruguay.py). Will be added in the next update.

Thanks

PR #985, opened 2021-04-09
Updated India count till 8th April 2021

abhimadhani:


edomt:
Thank you @abhimadhani - our updates for India are already automated, so this'll be added automatically.

PR #988, opened 2021-04-09
Update Cyprus.csv

twobytescy:


lucasrodes:
Thanks.

Added last row for April 7th,  as article states: 

> As of April 7th, a total of 174,626 vaccinations have been administered, of which 124,165 concern the administration of the 1st dose, and 50,461 concern the administration of both doses of the vaccine. The 1st dose of the vaccine has so far been administered to 9,788 individuals who fall into the vulnerable groups of the population.

PR #1000, opened 2021-04-10
Update 08-04

mouhib876:
Tunisia 124779 dose administrated 
Totally vaccinated:4860
https://www.facebook.com/186480324724413/posts/4023443124361428/


edomt:
Thanks

PR #1006, opened 2021-04-10
Update Bangladesh.csv

Shapod:


lucasrodes:
Thanks.

For the record: Might be soon suitable for batch automation with source http://103.247.238.92/webportal/pages/covid19-vaccination.php, once 2nd dose information is available.

PR #1031, opened 2021-04-14
Update Macao.csv

jesselau211:
this is the update for 12th and 13th of April


edomt:
@lucasrodes Seems like the script is not catching these unfortunately.

PR #1037, opened 2021-04-14
Update Bolivia.csv

mmorenoc25:
Vaccination update against COVID-19 in the Plurinational State of Bolivia corresponding to April 12 and 13, 2021 reported by the Ministry of Health and Sports.


lucasrodes:
Thanks

PR #1042, opened 2021-04-14
ENH: Re-structure of vaccination scripts

lucasrodes:
Vaccination structure refactor.

- Scripts now are moved to `src/vax/`.
- Environment can be easily installed using `$ pip install -e .`.
- `automation_state.csv`, `output/`, `requirements.txt` moved to the vaccination root folder.
- Python part of the pipeline is now executed as `$ python src/vax`, for more info use `$ python src/vax --help`.


edomt:
LGTM!

PR #1044, opened 2021-04-14
Azerbaijan Vaccination Update April 14

TheRalacator:
https://koronavirusinfo.az/az/post/765


edomt:
@TheRalacator We are now collecting this data each morning ourselves, so there is no need to open new pull requests. Thank you!

PR #1060, opened 2021-04-16
Cambodia - vaccination update 15-Apr-21

VIthaya97:
Cambodia covid-19 vaccination update as at 15 April 2021:
**- Total doses administered: 1,508,813**
- One dose given: 1,223,322
- Two doses given: 285,491

Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/194089-2021-04-16-02-08-22.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/194059-2021-04-15-11-53-22.html


lucasrodes:
Thanks!

PR #1070, opened 2021-04-16
Update Bangladesh.csv

Shapod:
Provided states here is not correct.Here 1st doses+2nd doses+doses applied to foreign nationals counted as a single dose ,hence numbers are different than official press release.

http://103.247.238.92/webportal/pages/covid19-vaccination.php


lucasrodes:
Thanks @Shapod

PR #1091, opened 2021-04-19
Update Hungary.csv

bdaniel88:


edomt:
Hi @bdaniel88 
Thanks :) 
We've changed things around a little bit, so the new file to edit is here: https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/output/Hungary.csv
I'll add this data point manually today 👍

PR #1095, opened 2021-04-19
Cambodia - Vaccination update 18-Apr-21

VIthaya97:
Sources:
MoH (from 16 April 2021): http://www.freshnewsasia.com/index.php/en/localnews/194216-2021-04-16-15-54-58.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/194383-2021-04-18-11-11-04.html


edomt:
Thanks

PR #1116, opened 2021-04-20
Update Curacao.csv

Djokison:
I have grabbed the (nearly daily) information on vaccinations published by the government of Curacao and added it. Data is accurate on the noted date at 10:00AM local time.

I have also removed the 2021-01-18 data point, as Curacao did not start vaccinating yet at that point; and have changed 2021-04-19, to follow the 10AM format.


Djokison:
Just saw that pull requests arent possible for the data.

lucasrodes:
Thanks for your contribution, we will be adding past data to Curaçao

PR #1124, opened 2021-04-21
Cambodia - vaccination update 20-Apr-21

VIthaya97:
Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/194686-2021-04-20-15-22-14.html
MoD:http://www.freshnewsasia.com/index.php/en/localnews/194680-2021-04-20-13-07-52.html


edomt:
Thanks

PR #1130, opened 2021-04-21
Add daily total incremental for Netherlands

thiagocotting:
Currently the vaccination data is incremented only weekly, since the source that's used now is updated every Tuesday. I want to add a new source for the total vaccination for the Netherlands to be able to follow it more accurately.


lucasrodes:
Added some changes, see #1132

PR #1131, opened 2021-04-21
Update Cyprus.csv

twobytescy:


lucasrodes:
Thanks

---

- Correct date: 2021-04-19
- Correct source url: https://www.pio.gov.cy/en/press-releases-article.html?id=19775#flat

PR #1134, opened 2021-04-22
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to March 30 to April 21, 2021.I strongly request that it be approved as I submit it to maintain the integrity of the information on Google and our word in data. The second doses come from the official site https://vacunas.panamasolidario.gob.pa.

### DIFF
--- a/public/data/vaccinations/country_data/Panama.csv
+++ b/public/data/vaccinations/country_data/Panama.csv
@@ -56,20 +56,24 @@ Panama,2021-03-27,Pfizer/BioNTech,https://twitter.com/MINSAPma/status/1375970370
 Panama,2021-03-28,Pfizer/BioNTech,https://twitter.com/MINSAPma/status/1376326017917333507,357431,,
 Panama,2021-03-29,Pfizer/BioNTech,https://twitter.com/MINSAPma/status/1376662708737675300,358277,,
 Panama,2021-03-30,Pfizer/BioNTech,https://fb.watch/4zFbSC-ruH/,364079,247539,116540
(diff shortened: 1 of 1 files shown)


lucasrodes:
Hi @mmorenoc25,
Thanks for your contributions, they are really helpful for this work. Some notes:

- Content in the folder [public](https://github.com/owid/covid-19-data/tree/master/public) is overwritten every day by our automation scripts.
- Currently, due to the nature of how the pipeline is built, we only are accepting PRs for country automation. We welcome your contribution for Panamá, the dashboard could potentially be scrapped in order to get the necessary data. You can find more details in #230 and #250.
- We only collect data whenever it is explicitly published. Hence, in the case of second doses, we would leave the field empty rather than fill it with last available value. This is because we don't really know the value for that day, and it might not be the same as the last value available.

Thanks once again for your help, feel free to ping me back if you have any questions/suggestions.

lucasrodes:
Track this issue in #1144

mmorenoc25:
Hello, very good evening from Panama, in view of all this, I considered that it would be better to eliminate the data of the second dose of which we do not have reliable and continue with the communications of the Ministry of Health are also published on the official website http://www.minsa.gob.pa, from what I see there is a difference of approximately 25,000 doses between both sources, why better to resort to the communiqués that inform the doses. However, it is very important that the second doses of the vaccinometer are not ignored, since it is the only official number that is had in that regard. Additional,I would like you to see what happens with the part of Google in which only 5 days of vaccination appear instead of those that previously appeared every day.

PR #1137, opened 2021-04-22
Update Hungary.csv

bdaniel88:


lucasrodes:
Thanks, should be added in today's automated data import.

PR #1141, opened 2021-04-22
VAX: Use a smaller source dataset in Czechia

kokes:
The Czech ministry has un-deprecated the region-based vaccination dataset, so we can use that one instead of the line-per-person one. This leads to a 100x decrease in bandwidth without losing any data necessary for our purposes.

```
output $ curl -I https://onemocneni-aktualne.mzcr.cz/api/v2/covid-19/ockovaci-mista.csv
...
content-length: 282840066
content-type: text/csv; charset=utf-8

output $ curl -I https://onemocneni-aktualne.mzcr.cz/api/v2/covid-19/ockovani.csv
...
content-length: 2366168
content-type: text/csv; charset=utf-8
```

The smaller dataset has a slightly different schema - the dose information is in columns, not rows, so we don't need to unstack. We also cannot `.size()` to get the total number of vaccines administered, we need to sum over `celkem_davek` (~total doses).

I verified the output using the current master and with my patch and got byte-for-byte identical results.

```
output $ shasum *.csv
d81a97c4870f5e6581521e49f9025083e48b9000  Czechia-master.csv
d81a97c4870f5e6581521e49f9025083e48b9000  Czechia.csv
output $ shasum by_manufacturer/*.csv
be39587f209f74e7b632d2af77d2c2d1ab35efd3  by_manufacturer/Czechia-master.csv
be39587f209f74e7b632d2af77d2c2d1ab35efd3  by_manufacturer/Czechia.csv
```

### DIFF
--- a/scripts/scripts/vaccinations/src/vax/batch/czechia.py
+++ b/scripts/scripts/vaccinations/src/vax/batch/czechia.py
@@ -26,10 +26,10 @@ def check_columns(input: pd.DataFrame) -> pd.DataFrame:
         "vakcina",
         "kraj_nuts_kod",
         "kraj_nazev",
-        "zarizeni_kod",
-        "zarizeni_nazev",
-        "poradi_davky",
         "vekova_skupina",
+        "prvnich_davek",
+        "druhych_davek",
+        "celkem_davek",
     ]
     if list(input.columns) != expected:
         raise ValueError(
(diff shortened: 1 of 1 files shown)


lucasrodes:
Thank you very much @kokes! That's a neat improvement

I took the opportunity to replace the usage of `input` reserved keyword for the more standardized `df`

lucasrodes:
LGTM

PR #1150, opened 2021-04-23
Update Cambodia.csv

VIthaya97:
Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/194956-2021-04-22-15-06-55.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/194936-2021-04-22-12-41-46.html


edomt:
Thank you!

PR #1156, opened 2021-04-23
VAX: Fix Thailand

petchumpriwan:
Update Thailand vax automation script regex due to change of report format


lucasrodes:
Thanks, @petchumpriwan!
I removed the exception catch, this way it becomes easier to detect that the pipeline is broken.

LGTM @edomt

petchumpriwan:
Thanks @lucasrodes, I saw the folder structure change quite a bit since my last contribution, so, now the pipeline is already handle error by itself right ?

lucasrodes:
When running the pipeline (`python src/vax`, read the [README.md](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/README.md) for more details), those scripts crashing (i.e. raising an exception) will be listed at the end.

Therefore, if we hide that exception raise behind a `print` statement that country won't be listed. You can see the logic in the following method: 

https://github.com/owid/covid-19-data/blob/e0e98bd44c7bdf2d3629fcb7645e8afa547b98f8/scripts/scripts/vaccinations/src/vax/__main__.py#L54

PR #1165, opened 2021-04-25
Update Panama.csv

mmorenoc25:
This is the last update of the vaccination process against COVID-19 in Panama, considering that we are using the information from the vaccine meter, I added the information from the previous days that appeared on the vaccine meter to have the most complete information. As reported by the government, the information in the statement and the report will be the same shortly, thank you very much for your attention, I appreciate it very much.


lucasrodes:
Thanks for your help and watch! We will be adding the new entries in today's update.

PR #1167, opened 2021-04-25
Update Cambodia.csv

VIthaya97:
Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/195233-2021-04-24-17-22-25.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/195229-2021-04-24-15-23-06.html


lucasrodes:
Thank you so much!

PR #1168, opened 2021-04-25
Update Thailand.csv

bombanoi:


lucasrodes:
Thanks, @bombanoi,
Data import for Thailand is automated. Before reporting new data points, you may prefer to wait at least 24 hours as not having the indicated data may mean that the automated import is broken.

PR #1170, opened 2021-04-26
Update Panama.csv

mmorenoc25:
Update of the vaccination against COVID-19 in the Republic of Panama corresponding to April 25, 2021 reported by the Ministry of Health and the Vacunometer of the National Government Innovation Authority (AIG)


edomt:
Thank you

PR #1173, opened 2021-04-26
Update 04-21-2021 Saint Kitts and Nevis.csv

Jaspac5:
Updated Vaccinations in Saint Kitts and Nevis as of April 21, 2021
Source:
https://www.facebook.com/StKittsHPU/posts/901952540588586


edomt:
Thanks

PR #1176, opened 2021-04-26
Update Cambodia.csv

VIthaya97:
Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/195331-2021-04-25-15-09-53.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/195322-2021-04-25-13-18-30.html


edomt:
Thank you!

PR #1177, opened 2021-04-26
VAX: Automate Philippines

edomt:
Work in progress - couldn't get the query to work for now.


edomt:
Thank you @lucasrodes!

PR #1195, opened 2021-04-28
Update Thailand.csv

bombanoi:


edomt:
Thank you - our data collection for Thailand is automated.

PR #1196, opened 2021-04-28
Update Panama.csv

mmorenoc25:
Complete update of the vaccination process against COVID-19 in the Republic of Panama from January 20 to April 27.The information of the second dose that appears in the vaccine meter historically from the day February 18. In order to verify this information you must enter the vaccinometer page https://vacunas.panamasolidario.gob.pa/vacunometro/, In the left tab, the calendar will appear which shows the number of doses supplied that day in a broken down manner, for which you must add one by one. I appreciate very much that the information is published as it is presented in full integrity. Thank you very much and good day.


edomt:
Thank you @mmorenoc25, we'll update our data accordingly.

PR #1208, opened 2021-04-28
VAX: Estonia batch

lucasrodes:
Using batch instead of Incremental


edomt:
@lucasrodes I'm not sure that I understand the idea behind `dt_limit = "2021-04-15"` — is it because of a 1-dose vaccine?

lucasrodes:
Indeed, I added this due to the _potential_ usage of 1-dose vaccine.

According to [the official gov site](https://vaktsineeri.ee/covid-19/vaktsineerimine-eestis/), a cargo of Janssen was received on the 13th of April. While this does not mean they started administering, I gave it the benefit of the doubt.

According to [their dashboard](https://www.terviseamet.ee/en/coronavirus-dataset) however, it seems that first doses + second doses = total_doses is still a valid relation. And no clear evidence that Janssen is being administered.

We can change the value, say `dt_limit = 3000-01-01`, for now. And revise in the future once there is evidence J&J is being administered

lucasrodes:
Updated the script to always perform 

```
total_vaccinations = people_vaccinated + people_fully_vaccinated
```

Screenshot of translated `VaccinationStatus` field in source data: 

image

Original can be found [here](https://www.terviseamet.ee/et/koroonaviirus/avaandmed#Covid-19%20vaktsineerimise%20%C3%BCleriigiline%20andmestik)

PR #1214, opened 2021-04-29
Update Japan.csv

poidowl:


edomt:
Thank you @poidowl, we'll fix this in the data.

PR #1217, opened 2021-04-29
Update Thailand.csv

bombanoi:


lucasrodes:
Note that public folder content is overwritten on every update. Feel free to issue PRs [here](https://github.com/owid/covid-19-data/tree/master/scripts/scripts/vaccinations/output) instead.

PR #1231, opened 2021-04-30
Update testing_data_corrections.R

camappel:
Remove UAE PR estimate

### DIFF
--- a/scripts/scripts/testing/testing_data_corrections.R
+++ b/scripts/scripts/testing/testing_data_corrections.R
@@ -10,3 +10,7 @@ collated[Country == "Ecuador" & Date < "2020-09-14", `Short-term tests per case`
 # Mauritania: the test definition does not match the case definition (screening tests possibly included in testing figures)
 collated[Country == "Mauritania", `Short-term positive rate` := NA]
 collated[Country == "Mauritania", `Short-term tests per case` := NA]
+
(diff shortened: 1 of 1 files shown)


camappel:
**UAE:** According to a [page](https://web.archive.org/web/20210430160249/https://u.ae/en/information-and-services/justice-safety-and-the-law/handling-the-covid-19-outbreak/types-of-tests-for-detecting-covid-19-and-testing-for-covid19) on the UAE's government portal, the UAE conducts multiple types of tests to detect COVID-19 in people, including antigen and antibody tests; it is not clear which types of tests are included in the testing figures. The case definition is also unclear, but a [circular](https://doh.gov.ae/-/media/597A7218DFB24A43976C40F571398086.ashx) published by the UAE Department of Health suggests that a positive PCR test is usually required to confirm cases of COVID-19 (although other types of tests may be used for screening or supporting diagnoses). Including non-diagnostic tests in testing figures may misrepresent testing capacity, and so we remove our estimation of the Positive Rate.

camappel:
**Iceland:** Testing figures include "Diagnostic tests", "Quarantine and random screening" tests, and "deCODE Genetics screening" tests, and exclude "Border screening 1 and 2" tests. The number of confirmed cases does not match the number of domestic infections from the types of tests included in the testing figures. We therefore remove our estimate of the Positive Rate.

PR #1239, opened 2021-05-01
VAX: Switch to official sources where possible for Bahamas

bardiharborow:


lucasrodes:
Thank you very much @bardiharborow for all the references.

We will be adding this for the next update.

Note that for manually updated countries our current pipeline does not accept PRs. More info [here](https://github.com/owid/covid-19-data/tree/master/scripts/scripts/vaccinations#faqs)

PR #1246, opened 2021-05-02
Automate Malaysia

jozhuatwx:
I'd like to highlight that there seems to be discrepancies between the [official website](https://www.vaksincovid.gov.my/statistik/) and the [official twitter](https://twitter.com/JKJAVMY). Not sure how you'd like to handle it. Additionally, I've never touched Python before this, I hope I've scripted this correctly. Thanks!


lucasrodes:
Thank you very much, @jozhuatwx.

I have added some minor corrections, but other than that, it LGTM! 😄 

As for the discrepancies, I have removed the last days as we do not allow for metrics to decrease, as they are cumulative. Let's hope that the data in both sources (both official) are aligned soon.

jozhuatwx:
Thanks for reviewing!

PR #1020, opened 2021-04-12
Cambodia - Vaccination update 12-Apr-21

VIthaya97:
Cambodia covid-19 vaccination update 12 April 2021:
- Total doses administered: 1,371,536
- One dose given: 1,099,811
- Two doses given (fully vaccinated): 271,725

Sources:
MoH: http://www.freshnewsasia.com/index.php/en/localnews/193533-2021-04-12-16-23-12.html
MoD: http://www.freshnewsasia.com/index.php/en/localnews/193511-2021-04-12-14-27-50.html

PR #1248, opened 2021-05-03
Update Cambodia.csv

VIthaya97:


lucasrodes:
Hi @VIthaya97 and thanks for your invaluable help.

Quick question, as it seems the format has slightly changed from previous press releases (https://github.com/owid/covid-19-data/pull/1020). I am trying to find the numbers that you provided in the table. Could you perhaps help me out with this?

In the table header, after some translation intents, I came to the conclusion that:

1: Cumulative number of people that got the first vaccine.
2: Cumulative number of people that got the second vaccine.
3: I am not sure here, I thought it was some breakdown (e.g. by Man/Woman), but the numbers in total don't match yours then. Maybe you can help me out with this

image

If these hypotheses are correct, I'd presume that we have:

1st doses: 1422722 (As you reported)
2nd doses: 984781 (slightly less than what you reported)

image


thanks in advance for you help!

VIthaya97:
Hi @lucasrodes,

Thanks for coming up with these questions.

You are definitely right. These are the columns that we need to look out for the cumulative numbers of one dose and two doses given (separated into total and females). 

I have seen a slight conflict made by both ministries (MoH and MoD) as their reports generated on that day were inconsistent. The good thing is the new report format by MoH will include the amount of vaccination by MoD as well (total vaccination across the whole country), but MoD will still publish their reports as usual. I have spotted a mistake made by the MoH report (the above report) that they included yesterday's figure of MoD instead of latest figure released by MoD. So, I personally did some calculation to correct this silly mistake. 

No worries, starting from today, I will report the numbers as they are shown on the table to avoid further referencing issues.

Thanks

PR #1249, opened 2021-05-03
Update Hungary.csv

bdaniel88:


lucasrodes:
Hi @bdaniel88,
As always, Thanks!

By the way, do you happen to know how Janssen doses are counted in these press releases?
Thanks

bdaniel88:
> Hi @bdaniel88,
> As always, Thanks!
> 
> By the way, do you happen to know how Janssen doses are counted in these press releases?
> Thanks

Hello, 

They just announced that we will start to use Janssen in a few days but I don't have any info about how they share the data about it. Well if I have any info, I'll inform u.
D

lucasrodes:
Thanks @bdaniel88!

PR #1253, opened 2021-05-04
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to April 30 and May 2 and 3. On the 30th, the data of the second dose of Vacunometer were added, On May 1 it was not reported and on May 2 and 3 it was reported so it is updated.


lucasrodes:
Thanks!
First dose for April not added,as  metric is not monotonically increasing

PR #1262, opened 2021-05-05
VAX : Automate New Caledonia

covid19owid:
Automate New Caledonia


edomt:
Hi @covid19owid 

Thank you for this, but in the end we've decided to import the data for New Caledonia as part of the data published by the SPC Public Health Division. See https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/src/vax/batch/spc.py

PR #1267, opened 2021-05-05
Update Dominican Republic.csv

mmorenoc25:
Vaccination update against COVID-19 in the Dominican Republic corresponding to May 3 reported by the Ministry of Public Health.


lucasrodes:
Note that Costa Rica has an automated process. The data should be added in today's automated import.

PR #1279, opened 2021-05-06
Update Vietnam.csv

Ltn123456:


lucasrodes:
Thanks @Ltn123456.

For Vietnam, we are now trying to source the data directly for the official source. 

It can be found at https://ncov.moh.gov.vn/, under "TIN TỨC" (navbar). Among these news, usually a daily COVID article is released before 7AM local time.

PR #1285, opened 2021-05-07
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to May 4, 5 and 6 reported by the Ministry of Health.


lucasrodes:
Hi @mmorenoc25,
Thanks. Note that the metrics need to be monotonically increasing. I have removed some values for `people_vaccinated`

PR #1288, opened 2021-05-07
Update Switzerland

antonelepfl:


antonelepfl:
I see that the process should be automated on this [table](https://github.com/owid/covid-19-data/blob/60801f98a1899d38e365d38a58422614769bdc80/scripts/scripts/vaccinations/automation_state.csv#L84) but last update was 5 days ago.

Not very sure why the table of Switzerland change the numbers that much but I just ran the script to fetch this information.

lucasrodes:
Hi @antonelepfl,
Thanks for reporting. Sometimes for Switzerland, the data is updated in the source files with a bit of delay, so no updates appear for some days.

In today's automated run, the data has been updated.

PR #1290, opened 2021-05-08
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to April 30, and from May 2 to 7 with the correction of the first dose so that its increase is the the right one. I appreciate the correction. 😎🇵🇦


lucasrodes:
Thanks @mmorenoc25 !

PR #1292, opened 2021-05-08
Update Namibia.csv

lihanh:
I've added missing data for the past week. All with verifiable sources.


lucasrodes:
Thank you very much!

PR #1295, opened 2021-05-08
Fix swapped values for Malta's vaccination data

ema-fox:
Malta's values for people_vaccinated and people_fully_vaccinated were swapped for 2021-05-06


lucasrodes:
We source our data from the [Official Open Dataset on COVID19 in Malta from the COVID-19 Public Health Response Team (Ministry for Health)](https://github.com/COVID19-Malta/COVID19-Cases)

I open the issue https://github.com/COVID19-Malta/COVID19-Cases/issues/5 to track this.

In the meantime I'd suggest modifying [script](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/src/vax/batch/malta.py) importing this data, so it does not import values after the 5th of May.

lucasrodes:
Moving this thread to #1310

PR #1307, opened 2021-05-10
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to May 9 reported by the Ministry of Health. I would like to check if all the numbers add up. Thank ✌🏽


edomt:
Thanks!

PR #1334, opened 2021-05-13
Update Cambodia - Past data update

VIthaya97:
Corrected wrong data entrances in Feb and added data for 17 March.


edomt:
Thanks

PR #1335, opened 2021-05-13
Update Hungary.csv

bdaniel88:
2nd dose corrected by the GOV today - but no announcement about it, so it's lower than yesterday :)


edomt:
Thank you! :)

PR #1338, opened 2021-05-13
Latest data change in Mexico

lumiroga:
-Fully vaccinated people wasn't displayed in exact number, was just talked about


lucasrodes:
Thanks!

PR #1346, opened 2021-05-15
Create China.csv

ira921208:
Corrected the wrong link and number.


lucasrodes:
Thanks for spotting, being fixed in the next update.

PR #1348, opened 2021-05-16
Update Panama.csv

mmorenoc25:
Vaccination update against COVID-19 in the Republic of Panama corresponding to May 13, 14 and 15.


lucasrodes:
Thanks!

PR #1350, opened 2021-05-16
Update Vietnam.csv

Ltn123456:
Update for 15/05. I also think that we should not use the data about 9/5 total vaccinations from WHO for consistency. So far, we have only used data from MoH. There is only a 4k difference between 888658 ( 9 May from WHO) and 892454 (10 May from MoH), (about 4,000 people vaccinated a day?) Furthermore, the number of those fully vaccinated (37k) is higher than that reported by MoH on 16/05 (https://ncov.moh.gov.vn/web/guest/-/6847426-3537). Both figures (total vaccinations from WHO and people vaccinated from MoH) referred to 9 May but I think the number from WHO was referencing a later data report (or different timezone?). MoH numbers are usually updated daily at 5-7 a.m,(UTC +7), with data current as of 4 p.m.(UTC +7) the previous day, while the other does not have any further information.


lucasrodes:
Thanks!

PR #1353, opened 2021-05-17
Update Panama.csv

mmorenoc25:
Update on vaccination against covid-19 in Panama. I added days and corrected first doses


edomt:
Thanks

PR #1387, opened 2021-05-20
Update Serbia.csv

dimadin:
This PR adds vaccination data for Serbia for May 18th 2021 which was not included. Since I feared that you might not include it, I have created a snapshot this morning when you could still see this data: http://archive.today/ehF8Z See also [Google cache](https://webcache.googleusercontent.com/search?q=cache:q0T3INsrO_IJ:https://vakcinacija.gov.rs/+&cd=1&hl=en&ct=clnk&gl=rs) (its [snapshot](http://archive.today/ybfHf)) for verification.


lucasrodes:
Thanks, @dimadin!

PR #1390, opened 2021-05-21
Update Cambodia.csv

davidc8:
Is it possible to automate data from pictures or images? The format published has been in consistency for a while now. I think we can work on automation for Cambodia. One thing I've noticed is the publishers (both official MoH Facebook page and Fresh News) always publish in images with no descriptions. Not sure if there is a tool that can compress and read images and retrieve data.


lucasrodes:
Hi @davidc8,
Thanks for your contribution!


On your questions: In the case of Cambodia there are two challenges:
- Getting an image from a social network: Need for Social Network API which means Token authorisation.
- Extract numbers from an image.

This might require substantial work and might not be very robust to format changes. However, if you feel confident in investing time on this, we'd be happy to review the outcome!

davidc8:
Hi @lucasrodes,

Thank you! I honestly do not think I can do this and it’s funny that i’m not even specialized and skilled in writing codes. I’m just trying to help translate available data for OWID. I was thinking that I would get help from you and the team but given that there would be a substantial work needed which is not worth it so, I’d rather keep on updating manually for now. If you would like to help on this some day in the future, i’m happy to assist. Thanks again!

PR #1394, opened 2021-05-21
Update Hungary.csv

bdaniel88:


lucasrodes:
Thanks, @bdaniel88 !

Do you know about any news in regards to Johnson&Johnson? Also, it seems like the single-shot CanSino vaccine might[ be on its way](https://www.globaltimes.cn/page/202105/1224162.shtml)

bdaniel88:
¡Hola! @lucasrodes ! The info I have about Cansino is that we're not planning to order it ( https://hungarytoday.hu/hungary-vaccination-covishield-cansino-vaccine-orders ) because of the late delivery dates and the government has enough in the storage/orders now. About Johnson&Johnson, a few days ago they corrected the numbers of second doses, maybe because they're realized that J&J is only 1 :). But I don't think they'll share anything about it, or how they'll publish the numbers. Unfortunately.

lucasrodes:
Köszönöm @bdaniel88 😄 !
By any chance, do you know how J&J is counted? If it  is included in first doses, second doses, ...?

We usually have  three metrics:
- `people vaccinated`: Number of persons that received at least one dose (either first dose for two-dose schemes or the only dose in single-shot schemes)
- `people fully vaccinated`: Number of persons  that have finished the vaccination scheme (two-dose and single-dose schemes)
- `total vaccinations`: Number of doses administered.

In the past, when no single dose was in use, the latter metric `total vaccinations` was obtained by adding up the other two (i.e. `people vaccinated` + `people fully vaccinated`). This is the way you are doing this at the moment. However, depending on how J&J is included in the metrics, this may have changed.

bdaniel88:
Unfortunately, they don't share any detailed info about it. Every day they publish a text about like this: "Vaccination is ongoing, with 4,969,321 people vaccinated, of whom 2,967,350 have already received their second vaccination."

PR #1401, opened 2021-05-22
Update Bahamas.csv

unguaio:


lucasrodes:
Thanks. Note that the number of total vaccinations and people vaccinated is not the same.

related #1385

PR #1402, opened 2021-05-22
Update Gambia.csv

unguaio:


lucasrodes:
Gambia is automated and we expect new data to be published soon. Lets keep an eye on this.

PR #1411, opened 2021-05-23
Fix Typo on Taiwan 2021-05-21 Vaccination Data

siwalan:
https://www.cdc.gov.tw/File/Get/a42bI1za87NU3AKC-MVCgA

Taiwan CDC only record 299,844 total vaccination on 21 May. In the previous (unfixed data), the git data said that the total vaccination is 2,998,442


lucasrodes:
Thanks! 

here were some checks to prevent this but were not raising an exception but a warning. Changed to Exception raising now to prevent this from happening!

Issue #1416, opened 2021-05-23
ENH: Add data broken by age group


Lokeshrathi:
If we could also add Age/Age-group in the dataset, that would work out great.
eg: 0-20,20-40,40-60,60-80

lucasrodes:
Hi @Lokeshrathi,
Thanks for your feature request. To better understand the proposal, if possible, could you elaborate a bit more on the following:

- Where would you add this data? That is, to which files?
- Which age ranges would you consider? 
- Do you have some examples of countries providing such data?

Thanks

PR #1434, opened 2021-05-25
Italy : Vaccination by age group

covid19owid:
Hi @lucasrodes ,

Vaccination by age group has been updated for Italy. Can you please review and provide your comments.

Thank you.


lucasrodes:
Hi @covid19owid,
Thanks for your work.

Prior to merging, I would suggest having a discussion (maybe in #1416).

My concerns are (each related to the previous):

- Get a sense of the countries that are sharing vaccination data by age.
- Related to the previous point, which age groups are typically present? For Italy, it seems it comes in slots of 10 years. Is that the norm?
- How can we ensure that comparison between countries is doable? That is, countries using different age groups?
- On the latter, maybe encoding the age group in two columns instead could help: `age_group_min` and `age_group_max` would help? Any potential drawback of this approach?

Thanks once again, and looking forward to adding this data.


cc @edomt

covid19owid:
Hi @lucasrodes ,

Thanks a lot for reviewing and great suggestions !

The output file has been updated to include age_group_min and age_group_max . The age groups differs for each country. So, it is better to use age_group_min and age_group_max columns for comparison. Please let me know your reviews.

Thank you.

lucasrodes:
Hi @covid19owid,
Thank you very much for your effort!

If you could share your thoughts on the other points, that would be really helpful!

>     * Get a sense of the countries that are sharing vaccination data by age.
> 
>     * Related to the previous point, which age groups are typically present? For Italy, it seems it comes in slots of 10 years. Is that the norm?
> 
>     * How can we ensure that comparison between countries is doable? That is, countries using different age groups?
>

covid19owid:
Hi @lucasrodes,

Thank you for the reviews. Please let me know your thoughts and comments for inline answers.

Get a sense of the countries that are sharing vaccination data by age.
Belgium, Canada, Chile, Czechia, Jersey, Lithuania, New Zealand, Norway are some of the countries which provide vaccination data by age. Will try to get list of other countries which provide vaccination by age.

Related to the previous point, which age groups are typically present? For Italy, it seems it comes in slots of 10 years. Is that the norm?
Age group 18+ is present. But  40+ is common. So, can we pick age groups which are common in the all countries.	 	 


How can we ensure that comparison between countries is doable? That is, countries using different age groups?
Age group is not even for all the countries. Some of the them are broken into 5 year slots. In this scenarios do we add up slots which are less than 10 to be in 10 years range. Ex : 40-45 and 46-50 becomes 40-50.

Or

Pick countries which have listed same age range vaccination data. Like Italy and New Zealand have same age range vaccination data. This should be the most likely approach now and also allows us to compare between countries.
        
Sample Source : 
https://raw.githubusercontent.com/italia/covid19-opendata-vaccini/master/dati/somministrazioni-vaccini-latest.csv
https://www.health.govt.nz/our-work/diseases-and-conditions/covid-19-novel-coronavirus/covid-19-data-and-statistics/covid-19-vaccine-data

lucasrodes:
Cool, Thanks, @covid19owid.

We'll add this data and wait to have more countries to generate a public dataset. This gives us some room to define further the file structure based on other countries.

scripts/scripts/vaccinations/src/vax/utils/paths.py
```diff
@@ -58,3 +58,6 @@ def tmp_vax_out(self, location):
 
     def tmp_vax_out_man(self, location):
         return os.path.join(self.tmp_vax_out_dir, "by_manufacturer", f"{location}.csv")
+
+    def tmp_vax_out_by_age_group(self, location):
```
lucasrodes, line 62: Great!

lucasrodes (approved):
LGTM

covid19owid:
Hi @lucasrodes,

Thanks a lot!. Will start creating separate PRs for other countries.

Thank you.

edomt (approved):

PR #1443, opened 2021-05-27
Update Vietnam.csv

davidc8:
The total vaccination is correct but spotted a bit of inconsistency for the 2nd dose administered. 2nd dose administered for the 2 days remain the same for the time for monotonic purpose.


edomt:
Thanks!

PR #1452, opened 2021-05-28
Remove Vaccination R scripts

lucasrodes:
R scripts have been deprecated by `cowid-vax generate` command.

They served great for almost 6 months. Their legacy will remain, now integrated into `cowid-vax`.


edomt:
RIP

PR #1460, opened 2021-05-30
Add six countries

sanjaygovindarajan:
Files tested on multiple dates to scrape data correctly. Files are not tested to update the spreadsheets correctly. 5 countries added:
-Antigua and Barbuda
-Saint Kitts and Nevis
-Equatorial Guinea
-Gabon
-Timor

Update 6/3:

Added Peru


lucasrodes (approved):
Thanks, @sanjaygovindarajan for your kind help!

I have added few changes to the structure. 

@camappel, I have reviewed the scripts and they (and the data they collect) look fine to me. If you could confirm that the data collected is correct and update our internal spreadsheet that'd be helpful. Thanks 😄 

---
Some minor comments:

- Just in case, force the date in the correct representation ("YYYY-mm-dd") and change the variable naming to avoid shadowing the `date` function. That is, use the following line instead: 
```python
date_str = date.today().strftime("%Y-%m-%d").
```
- Try to follow PEP 8 when it comes to spacing ([reference](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements)). Example:

✅ Correct:
```python
stats = soup.find_all('p', attrs={'class':'case-Number'})
```
❌ Wrong:
```python
stats = soup.find_all('p',attrs={'class':'case-Number'})
```
- Try to encapsulate the logic within a `main` function, and then call it from `if __name__=="__main__"`. See my changes for more details.

lucasrodes:
Hi @sanjaygovindarajan,
There is already a script for Peru in the batch folder.

camappel (approved):
These scripts all work properly and collect the relevant data.

Thank you very much for help again, @sanjaygovindarajan!

PR #1473, opened 2021-06-01
New Zealand : Vaccination by age group

covid19owid:
Vaccination by age group has been updated for New Zealand


lucasrodes:
Hi @covid19owid,
Thanks for your help!

Some notes:
- I have added and modified some code, trying to have it cleaner and more structured. For instance, only get the file link once (to the excel).
- I realised that while the data is batch, the age data is not. Note that the sheet containing it is overwritten each time. Therefore, we need to merge it with pre-existing. Otherwise, data is lost. I corrected it accordingly; see in method `to_csv`.

I will be working on some imports for European countries (age and manufacturer) in the upcoming days, so I'd suggest not to work on these to avoid any code conflicts.

Kind regards 😄

PR #1487, opened 2021-06-02
Uruguay: Output by vaccine manufacturer

3dgiordano:
Hi @edomt and @lucasrodes 

I had not realized that there was data by manufacturer.

I incorporated the generation of that output for Uruguay.

In Uruguay we have the data of vaccinated people and with full doses per vaccine.
I did not want to incorporate those columns because I noticed that they are not generating it for the countries.

It would be very easy if it was necessary to incorporate the columns people_vaccinated and people_fully_vaccinated

I have no experience in converting columns to rows, I used the technique of creating 3 dataframes and using concat to create the final dataframe.

Regards

### DIFF
--- a/scripts/scripts/vaccinations/src/vax/batch/uruguay.py
+++ b/scripts/scripts/vaccinations/src/vax/batch/uruguay.py
@@ -4,17 +4,49 @@ import pandas as pd
 def main(paths):
     df = pd.read_csv(
         "https://raw.githubusercontent.com/3dgiordano/covid-19-uy-vacc-data/main/data/Uruguay.csv",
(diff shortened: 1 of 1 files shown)


lucasrodes:
Hi,
Thanks for your help @3dgiordano!

I reviewed your code and it seems good! I substituted some of the code by using [pandas melt](https://pandas.pydata.org/docs/reference/api/pandas.melt.html) function.

As for now, we are only including numbers on total vaccinations, as it is easier to harmonize this metric among various countries. If we were to add other metrics, it is good to know that they'd be available for Uruguay, thanks for letting us know.

PR #1498, opened 2021-06-03
Add excess mortality + hospital + ICU to megafile & explorer

edomt:
@danielgavrilov I guess you'll also need it in one of the auxiliary files if we go ahead with this?


danielgavrilov (approved):
Thanks Ed! This is good to go on my side! 

Once live, we just need to add the new table and grapher configuration to the explorer.

edomt:
@danielgavrilov We also want hospital & ICU variables. Should I create another internal JSON or extend the excess mortality one? (does it matter?)

danielgavrilov:
@edomt I think there's already one there? https://github.com/owid/covid-19-data/blob/64ff03d41ba767571cac10818aaed1a04b865198/scripts/scripts/megafile.py#L421-L432

edomt:
Oh! All good then :)

PR #1503, opened 2021-06-04
Update Tunisia on 03/06

mouhib876:
Tunisia reache 1051796 dose administrated
First dose : 728092
Second dose: 322894
Also tunisia start using astrazenca vaccine 
This the link for the update of 03/06:
https://www.facebook.com/186480324724413/posts/4194043443968061/


lucasrodes:
Thanks!

PR #1520, opened 2021-06-08
Uruguay: By age group

3dgiordano:
Hi @lucasrodes 

I present a proposal that incorporates daily vaccination data by age from Uruguay following the new format indicated in the collaborative readme.
https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/CONTRIBUTE.md#age-group-data

The data are from the start of vaccination until 05/28/2021.
Access to daily data has been lost since that date, but work is ongoing.
There is another data source that is published approximately biweekly, which can be used to incorporate the most up-to-date data in the case that daily data is no longer available.

Regards
David

### DIFF
--- a/scripts/scripts/vaccinations/src/vax/batch/uruguay.py
+++ b/scripts/scripts/vaccinations/src/vax/batch/uruguay.py
@@ -1,4 +1,6 @@
+import re
 import pandas as pd
+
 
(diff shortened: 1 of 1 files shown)


lucasrodes:
Thanks @3dgiordano!

PR #1527, opened 2021-06-09
feat(yougov): add composite willingness to vaccinate dataset

bnjmacdonald:
Adds a new dataset (`YouGov-Imperial COVID-19 Behavior Tracker, composite variables.csv`), consisting of "composite" variables on willingness to vaccinate that are derived from a combination of YouGov survey data and official government data on vaccinations.

Dataset is used to create:
- https://ourworldindata.org/grapher/covid-vaccine-willingness-and-people-vaccinated-by-country
- https://ourworldindata.org/grapher/covid-vaccine-willingness-and-people-vaccinated-by-month


lucasrodes (approved):
Looks good to me!

I added slightly refactored yougov.py. Just minor things. Re-structured it as a class.

edomt (changes requested):
Not sure what's causing the issue, but no output is being generated at all.

lucasrodes:
There was a bug when exporting the file, thanks for catching it!

It seems to work, but I get error

```
Failed to construct composite variables.
```

When `DEBUG=True`

edomt:
> It seems to work, but I get the following error when `DEBUG=True`

@bnjmacdonald Let us know if this is something you want to fix now - otherwise we should be good to go.

edomt (commented):
@bnjmacdonald Let us know if this is something you want to fix now - otherwise we should be good to go.

edomt (approved):

bnjmacdonald:
Looks like the error with `debug=True` was b/c of an erroneous `df_temp.sample(10)` line I forgot to delete. Removed in latest commit and should now work with either `debug=False` or `debug=True`.

PR #1538, opened 2021-06-11
update taiwan.py to accomodate new PDF format

gugod:
On 2021/06/11, the first table inside the provided PDF is changed to
be a much smaller one. The accumulated total is still in the last row
but at different column.


lucasrodes:
Thanks @gugod !

PR #1606, opened 2021-06-25
Update Cambodia.csv

davidc8:


lucasrodes:
Thanks

davidc8:
Hi @lucasrodes,

Can you please revise the figure for 1st dose given? 

Seems like you inputted the number of people **intended** to get the vaccines instead of the number of people actually got vaccinated. The correct figure for 1st dose given should be 3,715,751. 

Thanks,

lucasrodes:
Thanks for the heads up, @davidc8. 

We've corrected it internally, will be live in the next update.

PR #1608, opened 2021-06-25
Add batch script for Japanese vaccinations.

fqj1994:
Japanese vaccinations data are usually backdated due to many reasons.

Government time-series data can provide more accurate information on
dose date, rather than taking the latest total doses.

This script also add support for per-manufacturer doses.


lucasrodes:
Hi @fqj1994,
Thanks for your help.

PR #1610, opened 2021-06-25
Uruguay: New age range '12_17'

3dgiordano:
Hi @lucasrodes 

The new age range 12 to 17 is incorporated.

Today I began to report that age range.
If the change is not incorporated, the process fails because it does not know how to map it.


lucasrodes:
Thanks @3dgiordano !

PR #1615, opened 2021-06-27
Patch Colombia VAX: Format Changed

pandres95:
Since a column for only doses was added to the format, this changed in consequence. This patch fixes the parsing of the data to match the new format.

Screenshot 2021-06-26 at 8 37 12 PM


lucasrodes:
Hi @pandres95,
Thanks for being on top of this.

From the introduced change, I suspect that a single-dose vaccine (J&J) might be about to start being administered (if not already)? 

This changes the interpretation of some metrics. In particular, the current computation of the metric `people_vaccinated` would not be correct and, perhaps, same for `people_fully_vaccinated`.

Do you have any idea on how single-doses (_únicas dosis_) will be reported? Ideally, the addition of a new cell _Total dosis aplicadas únicas dosis_ in the right would suffice.

PR #1618, opened 2021-06-27
feat(vax,japan): fix wrong vaccine type before Moderna was started.

fqj1994:
groupby added zero count group, which leads to "Moderna,
Pfizer/BioNTech" in vaccine column before Moderna was started.


lucasrodes:
Good catch. Thanks @fqj1994

PR #1621, opened 2021-06-28
Update Denmark.csv

unguaio:
As of 20 June, as reported here: https://covid19.who.int/region/euro/country/dk


lucasrodes:
Data for Denmark is automated using batch import, hence changes to the CSV files would be overwritten in future data updates. More details in our contribution guideline.

We are working to have Denmark data automated soon.

PR #1622, opened 2021-06-28
Update Hungary.csv

unguaio:
As of 20 June, as reported here: https://covid19.who.int/region/euro/country/hu


lucasrodes:
Thanks for reporting. 

We have added the data. Note that in order to accept future PRs, the modified file should be the one in [this folder](https://github.com/owid/covid-19-data/tree/master/scripts/scripts/vaccinations/output)

PR #1623, opened 2021-06-28
Update Papua New Guinea.csv

unguaio:
Automated data here: https://stats.pacificdata.org/vis?tm=covid&pg=0&df[ds]=SPC2&df[id]=DF_COVID_VACCINATION&df[ag]=SPC&df[vs]=1.0&pd=2021-02-02%2C&dq=D..&ly[cl]=INDICATOR&ly[rw]=TIME_PERIOD


lucasrodes:
Thanks for the report.

PR #1627, opened 2021-06-29
Update South Korea.csv

daehyeokbae:
Some Vaccination Data are incorrect and missing. Data Update.


edomt:
Thank you @daehyeokbae!
I've added the data for the missing days. However, the rest is not incorrect, it's based on another source (http://www.kdca.go.kr) that seems to have slightly different numbers, probably due to a different cutoff time.

PR #1639, opened 2021-06-30
Update Malaysia script

jozhuatwx:
Updated to read batch data from official source: https://github.com/CITF-Malaysia/citf-public


lucasrodes:
Thank you very much for your work @jozhuatwx !

PR #1651, opened 2021-07-02
taiwan: improve the table-extractor

gugod:
It looks like the main table may looks the same but structually
different -- the one today (7/2) comes with one less rows and the
position of N/A cells are different then before.

The parser (_parse_stats) is rewritten so it does not pull number from
certain positions, but instead scans the header column for certain
keywords. N/A cells are dropped so this should handle several cases
even if the author of pdf shifts some numbers left and right.

----

Here's the new output:

---
```diff
diff --git a/scripts/scripts/vaccinations/output/Taiwan.csv b/scripts/scripts/vaccinations/output/Taiwan.csv
index 82081e48a..a892e3d89 100644
--- a/scripts/scripts/vaccinations/output/Taiwan.csv
+++ b/scripts/scripts/vaccinations/output/Taiwan.csv
@@ -81,3 +81,4 @@ Taiwan,2021-06-27,"Moderna, Oxford/AstraZeneca",https://www.cdc.gov.tw/Category/
 Taiwan,2021-06-28,"Moderna, Oxford/AstraZeneca",https://www.cdc.gov.tw/Category/Page/9jFXNbCe-sFK9EImRRi2Og,1964592,1926572,38020
 Taiwan,2021-06-29,"Moderna, Oxford/AstraZeneca",https://www.cdc.gov.tw/Category/Page/9jFXNbCe-sFK9EImRRi2Og,2002677,1962990,39687
 Taiwan,2021-06-30,"Moderna, Oxford/AstraZeneca",https://www.cdc.gov.tw/Category/Page/9jFXNbCe-sFK9EImRRi2Og,2043218,2001418,41800
+Taiwan,2021-07-01,"Moderna, Oxford/AstraZeneca",https://www.cdc.gov.tw/Category/Page/9jFXNbCe-sFK9EImRRi2Og,2087587,2044241,43346
```

### DIFF
--- a/scripts/scripts/vaccinations/src/vax/incremental/taiwan.py
+++ b/scripts/scripts/vaccinations/src/vax/incremental/taiwan.py
@@ -58,8 +58,29 @@ class Taiwan:
     def _parse_stats(self, df: pd.DataFrame) -> int:
         if df.shape[1] != 4 or df.iloc[0,0] != "廠牌" or df.iloc[0,1] != "劑次" or not (df.iloc[-1,0] == "總計" or df.iloc[-2,0] == "總計"):
             raise ValueError(f"Table 1: format has changed!")
-        num_dose1 = clean_count(df.iloc[-3, 3])
-        num_dose2 = clean_count(df.iloc[-1, 3])
+
+        # We expect that the total of dose1/dose2 are in the last few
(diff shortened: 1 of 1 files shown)


lucasrodes:
Thanks for being on top of this

PR #1662, opened 2021-07-05
Update Thailand.csv

zielu92:
I found that in a detailed report (can be found on the page https://ddc.moph.go.th/vaccine-covid19/diaryPresentMonth/06/10/2021) starting from 25th June, there is also Sinopharm vaccine. On the daily report, there is no information about it.


lucasrodes:
Thanks

PR #1667, opened 2021-07-05
Update Denmark.csv

unguaio:
Total administered doses here: https://covid19.who.int/region/euro/country/dk


lucasrodes:
Thanks, already fixed using current source.

PR #1671, opened 2021-07-06
Update Serbia.csv

dimadin:


lucasrodes:
Thanks

dimadin:
@lucasrodes I don't see that data from this PR was added, however I see that rough, rounded numbers from one TV interview were added in 71b29e2. I suggest that you revert 71b29e2 for Serbia and add data from this PR.

PR #1673, opened 2021-07-06
Update Chad.csv

unguaio:


lucasrodes:
The proposed entry is not valid. Data value does not correspond to the reported date. Please check the source data in detail https://covid19.who.int/who-data/vaccination-data.csv

PR #1688, opened 2021-07-08
Update Cambodia testing.csv

davidc8:
Not so sure why the automation seems not to work for the past few days. Probably because of a slight change in new wordings.


edomt:
Thank you!

PR #1691, opened 2021-07-09
Update India.csv

nikhilnaik5:
Please use https://www.mohfw.gov.in/ in future because MOHFW stands for Ministry of Health and Family Welfare which is the GOV of India website. They provide the most accurate numbers.


lucasrodes:
Hi @nikhilnaik5,
We have switched to https://www.mohfw.gov.in/ in https://github.com/owid/covid-19-data/commit/58a04618f8269c915d8e76e7d76974c85c504a7e.

Thanks for reporting this.

PR #1694, opened 2021-07-09
fix(vax, japan): detect xlsx URL from homepage html.

fqj1994:
It's changed again, to numeric file names.....
It's probably will keep changing... let's detect the URL from HTML with some simple regex.


lucasrodes:
Thanks!

PR #1697, opened 2021-07-09
Update North Macedonia.csv

dacopavlov:
new data


lucasrodes:
Hi @dacopavlov,
I've noticed that you use report_date - 1, why is that?

Thanks for your help, it is very much appreciated

dacopavlov:
Hi @lucasrodes,

The provided data in the report from the officials is actually from yesterday. So the date in the post on Twitter is actually the date when the post is made, but the data is from the day before. 
I have an error in the pull request, the last row should have date 2021-07-08.

lucasrodes:
Thanks, @dacopavlov.

Do you know if there is any source that mentions this?

We usually check the date from the report, not the tweet itself. In this case, the reporting date in the image and the tweet posting date match though.

dacopavlov:
http://zdravstvo.gov.mk/vchera-aplicirani-11-393-dozi-vkupno-vakcinirani-496-228-lica/

It's in Macedonian, but you can translate it. It's a post from the official site of the ministry of health.

lucasrodes:
Thanks, @dacopavlov!

Adding the data accordingly

dacopavlov:
@lucasrodes the people_vaccinated number is not correct, should be 408953.

Best regards,
Dragan

PR #1750, opened 2021-07-21
data(vax): Czechia: Moderna to be renamed to Spikevax

kokes:
The ministry has announced it will rename Moderna to Spikevax in its open data exports. This change should take place on Thursday (tomorrow), so I'm prepping this PR just to be safe - it should only be merged if the pipeline fails tomorrow.

Sources:
- https://translate.google.com/translate?sl=auto&tl=en&u=https://data.nzis.cz/news-detail/cs/38-zmena-nazvu-ockovaci-latky-moderna-na-spikevax/
- https://twitter.com/MartinKomenda1/status/1417533017954869249


edomt:
It has indeed changed this morning :) 
Thank you so much!

PR #1754, opened 2021-07-22
Vaccine update automation for Armenia

menoua:
Adds script to scan official Facebook page of Armenia's ministry of health, where weekly announcements are made. The information is also available on the ministry's website (www.moh.am), but it fetches the announcements from the Facebook page using javascript, so I went directly to the source.

I'm sure there's a better way to get information from Facebook (e.g., using Facebook's GraphAPI), but since it needed permissions to be set up, I used a scraper package (https://github.com/kevinzg/facebook-scraper) as a quick fix.

I could not get the cowid-vax command working to test the code through it, but I used "incremental/albania.py" as template and tested the code independently, so it should be plug-and-play if everything is set up right.

### DIFF
--- a/scripts/scripts/vaccinations/requirements.txt
+++ b/scripts/scripts/vaccinations/requirements.txt
@@ -18,3 +18,4 @@ tabula-py==2.2.0
 tweepy~=3.10.0
 uk-covid19~=1.2.0
 xlrd==2.0.1
+facebook-scraper==0.2.45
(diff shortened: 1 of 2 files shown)


edomt:
Thank you so much @menoua, this is very useful! I'll merge now.

cc @lucasrodes - potentially a scraping solution we could extend to other countries without having to use the Facebook API

PR #1756, opened 2021-07-22
Update chile.py

camappel:
data(testing): 'Positividad' stopped updating on July 9. This scripts collects PCR and antigen tests (since both are used diagnostically)


edomt:
Thanks @camappel !

PR #1765, opened 2021-07-24
Update Algeria.csv

abdelouahabb:
please update the URL, this is a political election link that has nothing to do with covid/vaccine


edomt:
Thank you, we'll fix this today

PR #1778, opened 2021-07-29
Fix Vaccine(s) used for Singapore

rctay:
While Sinovac is available, those vaccinated by it do not make up the vaccination tally. 

Source:
- Ministry of Health emailed statement was quoted https://www.nasdaq.com/articles/singapore-not-counting-sinovac-shots-in-covid-19-vaccination-tally-2021-07-07 
- https://youtu.be/nov_3jnLn2w?t=27


lucasrodes:
Thanks!

PR #1782, opened 2021-07-30
Update united_kingdom.py

camappel:
data(test): exclude Ag tests for N. Ireland, Scotland and Wales


camappel:
Current metric (newVirusTests) includes PCR and Ag tests for all 4 countries, even though England is the only country that uses Ag tests to confirm cases

edomt:
Thank you @camappel ! See https://owid.slack.com/archives/C025BDNS5AA/p1627979990004400 for python formatting & linting

Issue #1786, opened 2021-08-02
Mexico vaccination dates are one day ahead starting 2021-07-27


saulpila:
The dates starting in line `201` of Mexico's vaccination data are one day ahead:
https://github.com/owid/covid-19-data/blob/15799c60ee9ec77fe2b021634964a6598160a692/public/data/vaccinations/country_data/Mexico.csv#L199-L204

Notice that 2021-07-27 was mistakenly skipped in the first column (line `200`).  (Before that line, dates are ok)

Mexico releases vaccination data the day after. For example, in line `204` above, https://www.gob.mx/cms/uploads/attachment/file/658742/2021.07.31_17h00.pdf contains vaccination data for 2021-07-30, *not* 2021-07-31. (The date on the link is the date of report, which is one day after the date of the data)

Thank you so much! Your dataset is amazing!

edomt:
Thank you!

PR #1787, opened 2021-08-02
Correct dates

saulpila:
This would fix https://github.com/owid/covid-19-data/issues/1786


edomt:
Thank you @saulpila ! We'll fix this in our input data.

PR #1780, opened 2021-07-30
Add J&J to Cambodia

davidc8:
Article: https://www.khmertimeskh.com/50904818/500000-doses-of-johnson-johnson-vaccines-courtesy-of-the-united-states-through-covax-initiative-lands-in-cambodia/

Inoculation starting date is expected to be 31-Jul-21 or 01-Aug-21. I will add J&J to the 'vaccines used' right on the date that the first shot is given. As of now, we have only received the doses.

PR #1790, opened 2021-08-02
Update Cambodia.csv

davidc8:
First shots of J&J have been given today. I ran formulas as guided by @lucasrodes in #1780. 11,863 doses of J&J were given today. Let me know if the formulas are run correctly. These look good from my end.


lucasrodes:
Hi @davidc8,
Thank you very much for your report. I have checked the numbers and they seem good to me.

PR #1809, opened 2021-08-05
Uruguay Vaccination with Booster Dose

3dgiordano:
Hi @lucasrodes 

Suggestion of incorporation of the new field with the additional boost doses.

It is already operational.

Regards


edomt:
Thank you @3dgiordano

PR #1813, opened 2021-08-06
Update China.csv

pytlv:
Added Aug. 6th data for China.


edomt:
Hi @pytlv 

See #230 
> For countries that are already included in our data, most of our imports are automated and will collect the latest number when we update our dataset. If you don't see the latest number appear in our data, please wait at least 24 hours before suggesting it here.

PR #1835, opened 2021-08-10
Update Tajikistan.csv

shaunfrancis:
Vaccination update for Tajikistan up to 4th August


lucasrodes:
Awesome! Thank you very much

Issue #1841, opened 2021-08-11
VAX: Add support for booster doses in main dataset


lucasrodes:
Several countries have started administering booster doses to their population.

Our current approach is to add a new metric that counts these: `total_boosters`. This will measure the total number of booster doses administered.

PR #1842, opened 2021-08-11
feat(vax): add `total_boosters` metric

lucasrodes:
This PR adds the new metric `total_boosters` and its per-capita counterpart `total_boosters_per_hundred`.

xref #1841


edomt (approved):

edomt:
LGTM, ready for merge whenever you want @lucasrodes :)

Rabelaiss:
Hi, pardon me if this is not the right place to ask.

From what I understood `total_boosters_per_hundred` = `total_boosters` / `population`, so if a person can receive two or more boosters then it becomes difficult to understand what this percentage means. Moreover, when 4 doses (2 boosters per person) are administered then `max(total_boosters_per_hundred)` raises from 100% to 200%, with 5 doses (3 boosters per person) it raises to 300% and so on.
On the other hand, I don't know if adding a new metric for each new booster is a good idea.

What are your thought on this?
Many thanks for the huge work you are doing for the community!!!

lucasrodes:
Hi @Rabelaiss,
That is correct.

`total_boosters_per_hundred` measures the total number of _boosters_ administered per hundred people. This can indeed well exceed 100% if second, third, ... boosters are administered.

We do it like this because we have found this to be the metric that is easiest to find for most governments.

Rabelaiss:
Many thanks for your kind reply

PR #1850, opened 2021-08-13
Update Tunisia

mouhib876:
Total doses administrated: 3 970 115
First dose: 2 703 133
Fully vaccinated: 1 689 253
The the link for the data:
https://evax.tn/vaccinationOD.html
Also Tunisia start using sinopharm vaccine


lucasrodes:
We will add Sinopharm in future updates. Hopefully WHO starts reporting it, too.

We rely on publicly accessible resources for our data. As evax.tn is not public (it is only accessible from Tunisia) we can't currently use it. Hopefully, it will be open to the public abroad so we can use it. Currently, we typically rely on reports from WHO or https://www.facebook.com/santetunisie.rns.tn.

mouhib876:
We don't use Facebook no more for the update on vaccination we use only
evax.tn
I can send a screenshot from the site if you want but it's in french only

lucasrodes:
Thanks for the offer @mouhib876, 
However, we only make data public that is openly accessible anywhere. We have contacted them to see if they will release this for the rest of the World.

mouhib876:
Hi I wrote the update for Tunisia 23 hours ago and it's not closed yet !!!

PR #1852, opened 2021-08-14
Update Cambodia & fix J&J duplication.csv

davidc8:
In the Cambodian Ministry of Health's announcement, both 1st doses (people vaccinated) and 2nd doses (completed dose) administered include J&J vaccine. So, total vaccinations must deduct J&J. Likewise, I found an error in 12 Aug 2021. This should be good now.


lucasrodes:
H @davidc8! Thanks for being so attentive. You are extremely helpful.

Kind regards

PR #1858, opened 2021-08-16
Add script for South Korean vaccination by manufacturer.

minyoh:
Current source of South korean vaccination offers per-manufacturer doses
so added some script to export per-manufacturer data.


lucasrodes:
Thanks, @minyoh.

Manufacturer vaccination data will be live from the next update.

Issue #1791, opened 2021-08-02
France: different count for vaccination 


B0pol:
The count for both 1st doses and fully vaccinated people is different between OWID and health autority data.

On 2021-07-29

- [OWID data](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/vaccinations/output/France.csv): 41673008 1st doses, 31841857 people fully vaccinated
- [Santé Public France data](https://www.data.gouv.fr/fr/datasets/r/fa4ad329-14ec-4394-85a4-c5df33769dff), from [data.gouv.fr](https://www.data.gouv.fr/fr/datasets/donnees-relatives-aux-personnes-vaccinees-contre-la-covid-19-1/): 41692674 1st doses, 34808758 people fully vactinated

Can you explain/fix?

edomt:
Dear @B0pol 

In our data, fully-vaccinated people are those who have received all doses prescribed by their vaccination regimen (e.g. 2 doses for Pfizer/BioNTech, Moderna, Oxford/AstraZeneca, etc. and 1 dose for Johnson & Johnson, CanSino, etc.).

Some countries also allow for alternative definitions, such as having been infected with SARS-CoV-2 in the past and having received 1 dose of a two-dose regimen. We currently ignore these alternative definitions to preserve the common definition of fully vaccinated, i.e. all doses required in the vaccine regimen. This allows for an optimal comparability between countries.

tristanguigue:
@edomt thanks for the explanation, I completely understand the need for simplicity however I would argue this does not necessarily allows for an optimal comparison between countries since those second doses will never be able to be administered given the local regulations. In particular, it will never be possible for a country such as France to have the fully vaccinated ratio equates the primary vaccination ratio which will be a problem as we approach the final stages of the vaccination in some countries. This might induce wrong conclusions to be drawn from this data, for example interpreting this as persons that refuse to get their second dose.

A possible solution could be that if and only if the fully vaccinated data is indeed publicly available from an official source then use it (such as is the case here).

edomt:
@tristanguigue That's a very interesting point, thank you for the feedback.

My personal hunch is that these alternative definitions might soon become obsolete. They were initially introduced in a context where there was a lock of vaccine doses, but I wouldn't be surprised if some of those governments announced in a little while that everyone is required to get the prescribed two doses, due to the waning effects of the immunity created by the disease.

[...]

PR #1866, opened 2021-08-18
Update Tunisia 17/08

mouhib876:
Total doses administrated: 4 737 282
First dose: 3 427 863
Fully vaccinated: 1 872 221 
Ps: fully vaccinated are calculated like this:
Second dose of vaccines + 1 dose of one shot vaccines + 1 dose with covid for people who had covid before vaccination


lucasrodes:
Hi @mouhib876,
Thanks for your report.

As per the calculation of `fully vaccinated`, we are currently sticking to the "traditional" interpretation (two doses required). You can read more on this in #1791

mouhib876:
Ok I get it then the number of fully vaccinated in Tunisia will be: 1 599 314

PR #1879, opened 2021-08-20
Update Serbia.csv

dimadin:
This PR makes two changes:

1. It adds data for August 19.
2. It adds numbers of administered booster doses.

On August 17th, Serbia [started administering](https://www.srbija.gov.rs/vest/en/177106/third-dose-of-covid-19-vaccine-available-as-of-tomorrow.php) third (booster) dose. Even though number of administered booster doses is available on [vakcinacija.gov.rs](https://vakcinacija.gov.rs/), it is not in a dataset. For August 17 and August 18, I retrieved those numbers by checking search engines caches. Thankfully, they were available, so I snapshoted them for a reference: https://archive.is/JoWuK and https://archive.is/Ej7eU


edomt:
Thank you!

PR #1881, opened 2021-08-20
Update thailand.py

camappel:
data(tests): fix Thailand automation and update automated_sheets


edomt:
Thank you @camappel ! :)

PR #1893, opened 2021-08-24
Update Uganda.csv

unguaio:
Also you can look here https://africacdc.org/covid-19-vaccination/


edomt:
Hi @unguaio 
We already get automated updates for Uganda via the WHO, with a full breakdown of people_vaccinated and people_fully_vaccinated (which the Africa CDC and the official Twitter account don't provide).

PR #1934, opened 2021-09-07
Update automation_state.csv

APinfinite:
https://github.com/owid/covid-19-data
``Mexico,True 
North Macedonia,True 
Palestine,True 
Saint Helena,True 
Saint Vincent and the Grenadines,true
Seychelles,true
South Africa,true
Tunisia,true
Uzbekistan,true``


lucasrodes:
Hi,
The processes for these countries are not automated, so their value should be `False`.

PR #1980, opened 2021-09-18
Update Cambodia.csv

davidc8:
Fixed wrong data for booster doses on 17th and updated for 18th

Edit: added new data for 19th, too.


camappel:
Thanks!

PR #1988, opened 2021-09-21
Update tunisia

mouhib876:
Total doses administrated: 7 249 413
First dose: 4 713 020
Fully vaccinated: 3 286 497


camappel:
Thanks!

PR #1991, opened 2021-09-22
Update Central African Republic.csv

unguaio:
Hello! I found out some problem of missing automated updates, I think due to some kind of wrong coordination referring to countries with double automated system like most african countries


edomt:
Hi @unguaio 

This is not a missed update: in your new row, `people_fully_vaccinated` is greater than `people_vaccinated`, which is impossible. That is what I already explained a few days ago.

PR #1992, opened 2021-09-22
Update Liberia.csv

unguaio:


edomt:
Hi @unguaio 

This is not a missed update: in your new row, `people_fully_vaccinated` is greater than `people_vaccinated`, which is impossible. That is what I already explained a few days ago.

PR #1995, opened 2021-09-22
Update Papua New Guinea.csv

unguaio:


edomt:
Hi @unguaio 

This data point does not exist in the SPC data.

PR #1998, opened 2021-09-23
Fix Taiwan: add Pfizer/BioNTech vaccine

gugod:
Since today (2021/09/23) the daily pdf start to include stats for Pfizer/BioNTech vaccine which is labeled as "BioNTech".


edomt:
Thank you!

PR #1999, opened 2021-09-23
Create spain.R

camappel:
data(tests): Script to automate Spain testing data collection. NOTE: This script only appends the provisional figure for the most recent date; it does not revise the consolidated figure for the previous week. If we can't automate a workaround I could manually revise the figure each week.


camappel:
Closed for maintenance

PR #2000, opened 2021-09-24
feat(vax,ecuador): fix vaccine mapping Cansino -> CanSino.

andrasio:
Expanding on #1976 

The vaccine `Cansino` is now being saved in our records as `CanSino`. For this reason we are opening this pull request so that the batch program doesn't try to rename the former name `Cansino`.


edomt:
Thank you!

PR #2001, opened 2021-09-25
Update Philippines.csv

moutmout00:
Correction of fully vaccinated number for 2021-09-23


edomt:
Thanks for flagging this! It'll be fixed in a few minutes.

PR #2007, opened 2021-09-25
Vaccination : Update Benin

covid19owid:
Update Benin

Source : https://new.in-24.com/world/114095.html


edomt:
Hi @covid19owid 

Thank you very much for all your contributions today! :) 

Just FYI, I'm going to be closing just a few of them without merging: for countries like Benin for example, we get our observations from the WHO, and they also include the list of vaccines used. It's very possible that your change is correct, and that the WHO data itself is out of date when it comes to the list of vaccines used. But as long as they don't update it, the next time we collect from them, the list of vaccines will go back to what it previously was (`"Oxford/AstraZeneca, Sinovac"`) because of the automated collection, so for now there's no point in changing the list.

PR #2027, opened 2021-09-28
Update tunisia

mouhib876:
Total doses administrated: 7 975 573
First dose: 5 014 695
Fully vaccinated: 3 768 319


edomt:
Hi @mouhib876 
For countries that are already included in our data, most of our imports are automated and will collect the latest number when we update our dataset. If you don't see the latest number appear in our data, please wait at least 24 hours before suggesting it here.

PR #2039, opened 2021-10-02
Testing : Update Andorra with Serology test

covid19owid:
Update Andorra with Serology test

Source : https://www.govern.ad/covid19/


edomt:
Thank you @covid19owid but we do not count serology tests in our testing data.

PR #2040, opened 2021-10-02
Testing : Update Jamaica

covid19owid:
Update Jamaica with Antigen testing

Source : https://www.moh.gov.jm/covid-19-clinical-management-summary-for-friday-october-1-2021/


edomt:
@covid19owid The `Testing type` variable in these files is actually deprecated; it's still there for some countries, but it doesn't do anything and doesn't appear anywhere in the final data.

PR #2106, opened 2021-11-01
enhance(vax,ukraine): batch update, manufacturer data

Golova1111:
Update Ukraine script (using batch update instead incremental), add Ukraine by manufacturer date

The previous script used scraped data from https://vaccination.covid19.gov.ua/, ("first doses" and "second doses" data), however, it leads to imprecise "full vaccinated data" as Ukraine started to use the J&J vaccine as well. 

I used data from the National Security and Defense Council of Ukraine portal instead (https://health-security.rnbo.gov.ua/vaccination) and fetch data from their API


scripts/src/cowidev/vax/batch/ukraine.py
```diff
@@ -0,0 +1,158 @@
+import datetime
+
+import pandas as pd
+import numpy as np
+import requests
+
+from scripts.src.cowidev.vax.utils.files import export_metadata
```
lucasrodes, line 7: Should be

```python
from cowidev.vax.utils.files import export_metadata
```

You can install the library by running `pip install -e .` from `/scripts/` folder

lucasrodes (commented):
Minor change required in the script. Other than that, looks good to me! Thanks for your contribution, @Golova1111.

scripts/src/cowidev/vax/batch/ukraine.py
Golova1111: thanks, updated

lucasrodes:
Thanks, @Golova1111. This is very helpful!
Ready to be merged

PR #2111, opened 2021-11-08
enhance(data): deprecate XLSX files on GitHub, moved to S3 #2082

lucasrodes:
Deprecation of XLSX files on GitHub as proposed in #2082.


edomt (approved):
Let's go! 🚮

PR #2116, opened 2021-11-10
Tunisia refactor

camappel:


scripts/scripts/testing/automations/incremental/tunisia.py
```diff
@@ -0,0 +1,42 @@
...
+    data = pd.read_csv(path).sort_values(by="Date", ascending=False)
+
+    source_url = "https://onmne.tn"
+
+    soup = get_soup(source_url)
+
+    cumulative_total = json.loads(soup.find("span", class_="vcex-milestone-time").attrs["data-options"])["endVal"]
+
+    Date = soup.select("p span")[0].text.replace("Chiffres clés mis à jour le ", "")
+    Date = pd.to_datetime(Date, format="%d %B %Y").strftime("%Y-%m-%d")
```
lucasrodes, line 23: The standard convention in Python is to use lower case, i.e. date instead of Date.

lucasrodes: Make use of library [extract_clean_date()](https://github.com/owid/covid-19-data/blob/a3a100d280ecbf0e5da0c654b4d09fae067da0a3/scripts/src/cowidev/utils/clean/dates.py#L66-L113):

```python
from cowidev.utils.clean import extract_clean_date
date = extract_clean_date(
    date_raw,
    regex=r"Chiffres clés mis à jour le (\d+ [A-Za-z]+ 20\d\d)",
    date_format="%d %B %Y",
    lang="fr"
)
```

Note that you can specify the language, so no need to use `locale.setlocale`. It is also safer to use the library function.


This will output the date in the format that we use, i.e. `YYYY-MM-DD`

lucasrodes (changes requested):

PR #2117, opened 2021-11-10
Kenya refactor

camappel:


scripts/scripts/testing/automations/incremental/kenya.py
```diff
@@ -0,0 +1,42 @@
...
+
+    soup = get_soup(source_url)
+
+    cumulative_total = int(
+        soup.find("div", class_="elementor-element-b36fad5").find(class_="elementor-text-editor").text.replace(",", "")
+    )
+
+    year = datetime.datetime.now().year
+    Date = soup.select(".elementor-element-75168b2 p")[0].text.replace("[Updated on ", "").replace(" [18:04]", "")
+    Date = str(year) + "-" + pd.to_datetime(Date, format="%B %d").strftime("%m-%d")
```
lucasrodes, line 23: You can use function [extract_clean_date()](https://github.com/owid/covid-19-data/blob/a3a100d280ecbf0e5da0c654b4d09fae067da0a3/scripts/src/cowidev/utils/clean/dates.py#L66-L113), which extracts the raw date using a regex expression and parses the date using the specified date format.

In this case we need to use function argument `replace_year` as the year is missing in the original text.


```python
from cowidev.utils.clean import extract_clean_date
date = extract_clean_date(
    date_raw,
    regex=r"\[Updated on ([A-Za-z]+ \d+) \[\d\d:\d\d\]",
    date_format="%B %d",
    replace_year=2021
)
```

Also, better to use `date` than `Date`, by convention variables in Python are lower case.

lucasrodes (changes requested):

lucasrodes (approved):

PR #2121, opened 2021-11-11
data(vax): manually update Taiwan.csv

gugod:
for some reason this was not included in the previous batch update.


lucasrodes (approved):
Thanks!

PR #2129, opened 2021-11-14
Testing: Belarus python port

bzki:
Related to: https://github.com/owid/covid-19-data/discussions/2099
I am able to successfully run `python3 automations/incremental/belarus.py`.  Would appreciate a code review and more rigorous testing as part of the various pipelines you run. 

3 notes:
- get_soup() has verify set to false (as it otherwise fails), and we receive a warning: `InsecureRequestWarning: Unverified HTTPS request is being made to host 'www.belarus.by'. Adding certificate verification is strongly advised`. 
- I did not use `from cowidev.utils.clean import extract_clean_date` for formatting the date, but let me know if that's an issue.
- Compared to the previous R script, empty values for the Notes column were `,"",` and are now `,,` since I am using pd.NA. Let me know if this is an issue or if there's a preferred way to handle this. 

Hopefully accepting this contribution is not more trouble than it is worth!


lucasrodes:
Just added some changes. Instead of using `extract_clean_date`, in your case we can use `clean_date` :).

Thank you very much for your contribution!

camappel:
The script runs perfectly, thanks!

PR #2138, opened 2021-11-19
vax; South Korea: Column names changed

minyoh:
Due to South Korean government's changes in terms and policies, Column names has been changed

- Changed term "일반접종(Ordinary/General dose)" to "기본접종(Primary/Basic dose)"
- New restriction prohibits vaccination of Moderna as primary dose to age under 30s


edomt:
Thank you very much @minyoh !

PR #2144, opened 2021-11-21
Update Tunisia

mouhib876:
Total doses administrated: 10 027 032
First dose: 5 900 488
Fully vaccinated: 4 856 968
Total boosters: 305 078


lucasrodes:
Thanks!

PR #2151, opened 2021-11-25
feat(vax,denmark): add total_boosters

edomt:
@lucasrodes Could you please debug this when you get a chance? The script is pretty long, and I struggle to understand what it does to get total_vaccinations – which seems to be where my new version crashes.
Thanks!


lucasrodes:
Done! Script for Denmark is probably too complex, should revisit at some point.

The mistake was coming from the fact that the script was trying to load the boosters file for very old dates, where no booster file was available. This is because `_parse_data` gets called for older dates (see the usage of `_get_total_vax`), to backfill the values for metric `total_vaccinations`. The booster info is not needed so I added the argument `load_boosters: bool` so we do not load boosters file in these cases.

Moreover, I added a final correction for `total_vaccinations`, to account for the number of boosters (see function `pipe_metrics`), as it only accounts for first and second doses.

lucasrodes (commented):
LGTM

edomt:
Thank you! ☺️

PR #2154, opened 2021-11-28
Uruguay, batch version for number of tests performed daily.

3dgiordano:
Hi @edomt and @lucasrodes 

I noticed that the data from Uruguay for the number of tests performed were being obtained incrementally from an html page.
The page was changed and that broke the data collection.

I propose a batch version.
They would have to evaluate removing the incremental version scheduled in R.

The data is from the CoronavirusUY application, which is an official application of the Ministry of Public Health of Uruguay.
But, the domain of this data is not hosted in gub.uy to circumvent access restrictions from abroad.

Check if the proposal is correct.
Locally I have run it, but since I don't have the csv to compare, I don't know if it is being generated in exactly the same format as the incremental one.

I hope it is useful.
Regards
David


camappel:
This works perfectly and matches our format. Thanks a lot @3dgiordano!

PR #2159, opened 2021-11-29
Fix taiwan pdf table format (changed since 2021/11/29)

gugod:
Hi,

The latest pdf files from 2021/11/29 is slightly changed. The embedded table becomes 3-column instead of 4, at least DataFrame parsed it so. (It still looks like a 4-column table though)

I'm not sure if such change is going to be permanent... it's probably better to support both new and old format for a while.

Please review this, thanks in advance!


edomt:
Thank you @gugod !

PR #2162, opened 2021-12-02
vax; South Korea: Column names changed

minyoh:
South Korean governmen has changed column names
- "기본접종(Primary/Basic dose)" to "1·2차 접종(1st·2nd dose)"
- "추가접종(Additional dose)" to "3차(부스터)접종(3rd(Booster) dose)"


edomt:
Thank you @minyoh !

PR #2170, opened 2021-12-05
fix(taiwan): the extraction of 2021-12-03

gugod:
Extra rows labled with "追加劑" are added to the table to show the
stats of 3rd doses -- but only under some of brands. I'm guessing we
will have to re-visit this parser soon.

### DIFF
--- a/scripts/src/cowidev/vax/incremental/taiwan.py
+++ b/scripts/src/cowidev/vax/incremental/taiwan.py
@@ -17,7 +17,7 @@ class Taiwan:
     def __init__(self):
         self.source_url = "https://www.cdc.gov.tw"
         self.location = "Taiwan"
-        self.columns_accepted = {"廠牌", "劑次", "12/2接種劑數", "累計至 12/2接種劑數"}
+        self.columns_accepted = {"廠牌", "劑次", "12/3接種劑數", "累計至 12/3接種劑數"}
         self.vaccines_mapping = {
             "AstraZeneca": "Oxford/AstraZeneca",
             "高端": "Medigen",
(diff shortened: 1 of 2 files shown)


scripts/src/cowidev/vax/incremental/taiwan.py
```diff
@@ -17,7 +17,7 @@ class Taiwan:
     def __init__(self):
         self.source_url = "https://www.cdc.gov.tw"
         self.location = "Taiwan"
-        self.columns_accepted = {"廠牌", "劑次", "12/2接種劑數", "累計至 12/2接種劑數"}
+        self.columns_accepted = {"廠牌", "劑次", "12/3接種劑數", "累計至 12/3接種劑數"}
```
gugod, line 20: The column heades comes with a date (usually yesterday) and we probably want to make this a regex so we don't have to change this every day...

scripts/src/cowidev/vax/incremental/taiwan.py
```diff
@@ -101,7 +102,7 @@ def _parse_stats(self, df: pd.DataFrame) -> int:
         #     num_dose1 = clean_count(num1.split(" ")[-1])
         #     num_dose2 = clean_count(num2.split(" ")[-1])
 
-        if df.shape != (11, 2):
+        if df.shape != (13, 2):
```
gugod, line 105: Assuming the stats for 3rd dose (labeled as "追加劑") will be included under all branchs... there will be two more rows and df.shape may become `(15,2)`

gugod:
I've fixed the parser in `taiwan.py` so it parses the stats of 12/3 -- but it'll likely fail again latter  when parsing the pdf released at 12/6 (Mon) because of those sanity checkes. I'll try to follow-up on recent changes as much as I can.

gugod:
Replaced by #2171

scripts/src/cowidev/vax/incremental/taiwan.py
gugod: The suffix of last 2 columns are latter changed to be: "接種人次" instead.

PR #2171, opened 2021-12-06
fix(vax,taiwan): pdf extraction for 2021-12-5  and 2021-12-3

gugod:
Due to recent format changes of newly add rows for 3rd doses some assumption about the shape of the forms are no longer true. They are adjusted accordingly.

This one includes  commits from  #2170 -- therefore I've closed #2170.

### DIFF
--- a/scripts/src/cowidev/vax/incremental/taiwan.py
+++ b/scripts/src/cowidev/vax/incremental/taiwan.py
@@ -17,7 +17,7 @@ class Taiwan:
     def __init__(self):
         self.source_url = "https://www.cdc.gov.tw"
         self.location = "Taiwan"
-        self.columns_accepted = {"廠牌", "劑次", "12/2接種劑數", "累計至 12/2接種劑數"}
+        self.columns_accepted = {"廠牌", "劑次", "12/3接種劑數", "累計至 12/3接種劑數"}
         self.vaccines_mapping = {
             "AstraZeneca": "Oxford/AstraZeneca",
             "高端": "Medigen",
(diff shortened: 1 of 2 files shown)


scripts/src/cowidev/vax/incremental/taiwan.py
```diff
@@ -51,13 +50,15 @@ def _parse_pdf_link(self, soup) -> str:
     def _parse_table(self, url_pdf: str):
         dfs = self._parse_tables_all(url_pdf)
         df = dfs[0]
+
         # Sanity check
-        cols_wrong = df.columns.difference(self.columns_accepted)
-        if cols_wrong.any():
-            raise ValueError(f"There are some unknown columns: {cols_wrong}")
+        cols = df.columns
+        if not (len(cols) == 4 and cols[0] == "廠牌" and cols[1] == "劑次" and cols[2].endswith("接種人次") and cols[3].startswith("累計") and cols[3].endswith("接種人次")):
```
lucasrodes, line 56: Here I would suggest using [the walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) and regular expression instead:

```python
# Sanity check
if not (
    len(cols := df.columns) == 4
    and cols[0] == "廠牌"
    and cols[1] == "劑次"
    and cols[2].endswith("接種人次")
    and re.match(r"\d+\/\d+\-\d+\/\d+接種人次", cols[2])
    and re.match(r"累計至 \d+\/\d+接種人次", cols[3])
):
```

lucasrodes:
Thanks for your contribution! Added some comments, I'll apply this after merging as (i) the code works and (ii) I do not have permissions to push to your fork.

scripts/src/cowidev/vax/incremental/taiwan.py
gugod: @lucasrodes thanks for the advice.

We might have to change the patterns of `cols[2]` from:

```
   and re.match(r"\d+\/\d+\-\d+\/\d+接種人次", cols[2])
```

to:

```
   and re.match(r"(\d+\/\d+\-)?\d+\/\d+接種人次", cols[2])
```

As the first date stamp is only there on Mondays.

Here's a repo with past PDFs for references: https://github.com/profmike/tw-vaccine-stats/tree/main/data/tw.gov.cdc.vaccination_reports/daily

PR #2179, opened 2021-12-07
fix(taiwan): update the sanity checks according to new format.

gugod:
The sanity checks of column names are relaxed a little bit to account
for whitespaces conventionally put in between digits and Chinese
characters.

Also, the number of rows has increased due to the newly added stats
for booster shots ("追加劑" in the table.)


edomt:
Thank you!

PR #2180, opened 2021-12-08
Taiwan csv: include totoal booster

gugod:
I've patch the incremental parser to include the columne "total_booster" -- which has been provided since 2021-12-02. This seemed to create a larg diff on the output.. I'm not sure if it is ok.

Meanwhile I've manually typed in the those numbers for the past few days (commit 32d4c11eb4040a71303cc670d4599313c6615820)


edomt:
That's great, thank you again!

PR #2182, opened 2021-12-13
Taiwan.csv: manual updates for 2021-12-12

gugod:
```
2021-12-14 00:30:03 ERROR    cowidev.vax.incremental.taiwan: ❌ There are some unknown columns: Index(['廠牌', '劑次', '12/11-12/12接種人次 累計至 12/12接種人次'], dtype='object')
```

The parser falied at such error -- which is caused by the fact that the source dataframe object `df` having a incorrate numebr of columens -- it become 3 instead of the usual 4 -- while the oveal table lookßlike it has 4 columns.

I'm not sure whether this is just a blip or if this is going to persist... so I decided to wait for a nother day before decding if a follow-up code changes is required.

Meanwhile, where's numbers that I manually typed in.


edomt:
Thank you for the update, and for monitoring this so closely! 🙏🏻

PR #2203, opened 2021-12-27
Refactor(test,italy) migrate to python

camappel:


lucasrodes:
Thanks, @camappel! 
Note in commit 298a064 how I define the output path. I am trying to make it work in multiple environments and hopefully, this approach can help.

PR #2212, opened 2022-01-01
Manually add the entry of 2021/12/31 to Taiwan.csv

gugod:
The PDF released on 2022/01/01 triggerd such error:

    ValueError: There are some unknown columns: Index(['廠牌 劑次', '110/12/31 接種人次', '累計至 110/12/31 接種人次'], dtype='object')

Similar to PR #2182, I'm guessing that this is due to a human error
that isn't likely to persist. Therefore I've only updated the CSV file
but not the parser.


lucasrodes:
Thank you very much, @gugod!

PR #2214, opened 2022-01-03
Manually patch one entry to Taiwan.csv

gugod:
The PDF released on 2022/01/03 still persists with the same error as

    ValueError: There are some unknown columns: Index(['廠牌 劑次', '110/12/31-111/1/2 接種人次', '累計至 111/1/2 接種人次'], dtype='object')

Which is the same as the one described in commit 67d5d3e1ab497d935c2d70da56aa0ed9cb4959bf

Previously: #2212, #2182


lucasrodes:
Thanks, @gugod. Hopefully, they will be fixing this soon.

PR #2215, opened 2022-01-04
update taiwan.py parser for year 2022.

gugod:
Happy new year :)

Luckly the errors in #2214 and #2212 did not persist. But we forgot to mark the date corrrectly so Taiwan.csv has not been updated eventhough all the numbers are extracted correctly.

Also I include a few changes that re-align first column so we don't take numbers from a 'NA' row with others as well cleaning up this FutureWaring, which I encountered when running with python3.9
 
    FutureWarning: The default value of regex will change from True to False in a future version


lucasrodes:
Thanks, @gugod,
LGTM

PR #2216, opened 2022-01-04
Update incremental.py

camappel:
Allow scripts to append daily change without cumulative total


camappel:
The problem was that the sanity check on line 84 stopped the script if cumulative_total == None. I changed it so that it only ran the cumulative_total checks if daily_change == None.

lucasrodes:
Thanks!

PR #2230, opened 2022-01-07
manual update of Taiwan.csv for 2022-01-05

gugod:
The same error shows up for another day. I shall spend some time to update the parser... but first let me commit the new stats.

previously #2226


edomt:
Thank you @gugod !

PR #2233, opened 2022-01-08
feat(hosp,japan): add Japan data

fqj1994:
Not sure if it is a good idea to crawl all historical weekly xlsx file, but hosp module doesn't seem to support incremental.

### DIFF
--- a/scripts/src/cowidev/hosp/sources/japan.py
+++ b/scripts/src/cowidev/hosp/sources/japan.py
@@ -1,0 +1,55 @@
+import datetime
+import io
+import re
+import sys
+
+import requests
+import pandas as pd
+
+DOMAIN = "https://www.mhlw.go.jp"
+
+METADATA = {
+    "source_url": DOMAIN + "/stf/seisakunitsuite/newpage_00023.html",
+    "source_url_ref": DOMAIN + "/stf/seisakunitsuite/newpage_00023.html",
+    "source_name": "Ministry of Health, Labour and Welfare",
+    "entity": "Japan",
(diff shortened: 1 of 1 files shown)


edomt:
Thank you very much @fqj1994 - ideally we'd like to have all the data coming from one source, so if you ever find such a source, please let us know!

But I've wanted to add hospital data for Japan for a while, so this is a good intermediate solution :)

fqj1994:
Hi,

NHK has a single source of aggregated data, but I don't think we should crawl from it since the license requirement (no sub-license allowed, detailed requirement on attributions) may not fit for owid's needs.

Another question is,
JP has a "critically ill-ed" patients data in the same xlsx, defined by in an ICU **or** on a ventilator.
In Asia, there was a long history that putting on a ventilator outside an ICU, even the situation has been improving these years. I'm not sure if COVID will make this situation better or worse. Also patient could be put into ICU for other reason or close monitoring during COVID treatment in an aging country.
I'm not sure this data fit OWID's standard for "ICU patients"? What do you think? Should we include it or just ignore it?

But anyway, even for "hospitalization" in JP, the criteria keeps changing, making the data less meaningful, and they tends to put more percentage of cases hospitalized when infections are low, to include mild cases with higher risk, and only trying to hospitalize medically necessary patients when infections are high. (for example, the central government just cancelled the policy to hospitalize all omicron cases early this week, and let local government to decide). It happened occasionally last year in some region that mild patient occupied the bed earlier pushes the limit of medical system to accept new serve patients.

PR #2235, opened 2022-01-08
manual update of Taiwan.csv for stats of 2022-01-06 and 2022-01-07

gugod:
another manual updates, before I figure out how to make the parser more robust. :)


lucasrodes:
Thanks, @gugod !

PR #2240, opened 2022-01-10
update taiwan.py and make it a bit more robust.

gugod:
One key-chainge is to enable "lattice mode" when invoking tabular.read_pdf()

Since it looks like all PDFs released by cdc.gov.tw come with grid
lines and by enabling lattice mode, read_pdf produce a 20x4 table
consistently for at least 7 latest PDFs. Although a few rows still
requires some tweaks, it is much easier to work with.

### DIFF
--- a/scripts/src/cowidev/vax/incremental/taiwan.py
+++ b/scripts/src/cowidev/vax/incremental/taiwan.py
@@ -1,5 +1,5 @@
 import re
-
+import math
 import pandas as pd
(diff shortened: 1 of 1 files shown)


lucasrodes:
Thanks for taking the time,
LGTM


PS. Hopefully, they will release a CSV-like file soon 😄?

lucasrodes (approved):

PR #2242, opened 2022-01-10
Update Armenia.csv

menoua:
New numbers for Jan 9. Also, updated WHO Jan 2 numbers by the ministry of health numbers (which are lower). Still don't know why there's a mismatch between the MoH and WHO numbers for Armenia.


lucasrodes:
Thanks!

PR #2251, opened 2022-01-11
data(tests): africa cdc scraper

camappel:
I made a start on a script to scrape testing data for African nations from the Africa CDC


lucasrodes:
@camappel 
I have added some changes:

1. Testing `AfricaCDC` object is using some logic from the vaccination module (class inheritance).
2. Added some preliminary list of countries to extract from Africa CDC
3. Added temporary function under `scripts/scripts/testing/automations/incremental/africacdc.py` which basically runs the code from the new script. This is basically temporary while we still use the current pipeline ([`run_python_scripts.py`](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/run_python_scripts.py))

I'd suggest you try to run the code

```
cd scripts/scripts/testing/automations/incremental/
python africacdc.py
```

And check the newly generated files.

Also, please feel free to edit the value of `ACDC_COUNTRIES` by adding or removing countries.

PR #2271, opened 2022-01-17
(vax, iran) source change

Rouein:
I changed the source url and made some changes accordingly. 
it should not have any failures.

PR #2272, opened 2022-01-17
(test, iran) python script

Rouein:
I created a python script to replace the one written in R for testing.


Rouein:
This pull request should not merge before the previous pull request has been merged!

lucasrodes:
Hi @Rouein!
Thanks for your time again. Could you reference the ID of "the previous pull request"? Thanks!

edomt:
> Hi @Rouein! Thanks for your time again. Could you reference the ID of "the previous pull request"? Thanks!

I think it's #2271

camappel:
Hi @Rouein,

This scripts works perfectly - thanks a lot!

Rouein:
you'r welcome!
I used the same structure we developed for vaccination.
It could be generalized more with variables for  html elements and could easily be applied to all news feed sources.(in case there is a need.)
If you are happy with the structure I can use it to translate a few more of the R written scripts.

lucasrodes:
Hi @Rouein,
Thank you so much for the proposal. We could try to have the generalization for news feed-like sources within `cowidev` library. This could potentially be a class, from which all countries needing incremental updates from news feed could inherit, or, alternatively, a function.

FYI, currently, the only generalization attempt for testing scripts can be found in [`cowidev.testing.utils.base`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/utils/base.py) module, where we have the class `CountryTestBase`, which currently only supports _batch_-like processes (like [these](https://github.com/owid/covid-19-data/tree/master/scripts/scripts/testing/automations/batch), [`cuba.py`](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/automations/batch/cuba.py) as an example).

PR #2279, opened 2022-01-18
Testing Mexico

3dgiordano:
Hi there.

I migrated the logic from the R code over testing data from Mexico.

I did not test the final code in the class format of CountryTestBase class.
Possibly it is already functional and generate the same result

Apologies if there are any minor errors.

PS: I fixed a bug in the Positive Rate calculation that left it NaN in the R version when it had no data to compute the value

cc. @camappel 

Regards
David


lucasrodes (approved):
Thanks 🙏. I fixed some minor issues due to desync with master.

PR #2281, opened 2022-01-18
Fix taiwan pdf table column name validation 20220118

gugod:
In #2275 I changed the column name validation to support format `yyy/mm/dd` since that the format used in the pdf released at 2022/01/17. However in the pdf released at 2022/01/18, the date formatin the table header is in `mm/dd`.

The regex is updated to support both formats.


lucasrodes:
Thanks, @gugod 🎉 
Looks as if there is a different person writing the PDF every day 😄

PR #2283, opened 2022-01-18
Spain testing data

3dgiordano:
Hi all

I noticed that you guys don't have a batch for Spain testing data.

I did one, it incorporates Antigens and PCR, I noticed that the classification of the type of test is not clear, or if they are preferring it to be only PCR.

It's easy just as you adapt it to your criteria.

As in previous collaborations, I did it not based on its class model, so I migrated it to its model without testing it.

I hope it is useful.

Regards
David


camappel:
Hi @3dgiordano,

Thank you for finding this source - it's actually preferable to our current one. I'm adapting your script now so that it runs accordingly.

3dgiordano:
Great @camappel !

Pleased to collaborate with your excellent work. 

Regards
David

PR #2287, opened 2022-01-20
util test increment

Rouein:
if only daily change is available


edomt:
Thank you, good point

Issue #2278, opened 2022-01-17
Georgia testing data cutoff


czka:
As of d2e597487d Georgia testing data in `owid-covid-data.csv` started on 2021-04-09. As of current master (a6f0f1ae5f writing this) they start on 2021-11-07. Were those records removed deliberately?

camappel:
Hello @czka 

Yes, the data for `cumulative tests` performed were removed, as we now collect data for the `daily change in the number of tests` performed. For the positive rate to appear in our [grapher](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=asc&pickerMetric=location&Metric=Share+of+positive+tests&Interval=7-day+rolling+average&Relative+to+Population=true&Color+by+test+positivity=false&country=~GEO), we can only have one metric in the csv, and so I made the decision to remove the `cumulative tests` figures, as I feel it is more important to have an up-to-date estimate of the positive rate.

Thanks a lot for your question!

camappel:
Got it, thanks

czka:
Hi,

Why was the issue closed? As of 1d90897f8b Georgia's testing data are still missing.

czka:
Hmm. I didn't mean cummulative test count. I meant `new_tests` and `new_tests_smoothed`. Georgia's records for these 2 fields in `owid-covid-data.csv` as of d2e5974 started on 2021-04-09. Now they start at 2021-11-07.

camappel:
I see. The figures for new_tests and new_tests_smoothed would have been generated based on the daily difference between the cumulative total. It would be possible to re-add those figures - thank you for pointing this out

camappel:
I just re-added the old data.

In the future, I would like to write a script that scrapes the [source](https://agenda.ge/en/news) for reports like [this](https://agenda.ge/en/news/2021/915) one, looking for some variation of '[x] tests have been conducted around the country in the past 24 hours'; this would allow to have a continuous series back to the start of the pandemic, thus calculating the cumulative total.

Rouein:
I will write an "incremental" scrapper for Georgia based on the [source](https://agenda.ge/en/news). I just need a start date, should I start from 2021-04-09?

Rouein:
ok. it might take about 24 hours if they don't block my IP

PR #2288, opened 2022-01-20
Albania test python script

Rouein:
#2287 is already implemented in this branch!


camappel (approved):
This works perfectly, and makes good use of the cowidev module. 

Next, we can work on scraping the missing data points.

Thanks!

Rouein:
the website doesn't have robots.txt and with delay I was being blocked every now and then but finally I managed to get all the missing data.

camappel:
Very impressive work, thank you!

If you feel like making another contribution, perhaps you could adapt the script you used to scrape the missing the data points to do the same for Georgia, as discussed in #2278 ?

Rouein:
can you be more specific? what data is missing for Georgia?

Rouein:
is there no automation script for georgia?

PR #2290, opened 2022-01-20
Finland use new data source for testing data

3dgiordano:
Hi all!

I discovered that they do not have the updated data.
I entered the page where they get the data and it indicates that it was replaced by a new page.
They also enabled a query api that allows you to query the necessary data and export it in csv.

I did a batch update to migrate to the new data source.

I clarify that I did not test the class, I migrated my batch I had to its class format but I could not test it.
Apologies if I incorporate a minor bug.

There is an issue about how Finland updates its data, the reports, since they update their data backwards.
It is possible that the data of the last date or the previous one has partial information and it is not convenient to publish it.
I leave it to your discretion to evaluate whether or not to exclude the data from those last days.
I think something similar already happened in the previous version, since it was a disclaimer also in the old data source.

I also noticed that the previous batch used a csv format oriented to accumulated tests, the new version has the format used by other countries since it has a number of tests and positive cases.

I hope it helps.

cc @camappel 

Regards
David


edomt (approved):

edomt:
All good for me, thank you @3dgiordano ! Over to @camappel to check the results, and potentially edit the documentation for Finland if needed.

camappel (approved):
Another great contribution - thanks!

PR #2305, opened 2022-01-22
Test moldova new script

Rouein:
python script for Moldova's test data(daily change) using the ministry's website as a source (the old source was not providing date of the data it had also changed the format from cumulative total to daily change)


Rouein:
since the new source works with a web driver scraping for old data will take so much time making it almost impossible.

Rouein:
spoke too soon!

lucasrodes (approved):
LGTM.

This PR changes from `Cumulative total` to `Daily change in cumulative total`.

The motivation for this is the availability of the metrics. It looks like daily updates on the latter are available.

Rouein:
Unfortunately, that's the most reliable source I could find(the governmental website), and still on some occasions the do not post the announcement containing the data.  So, for the sake of accuracy I had to stick with daily change in cumulative data.

camappel (approved):
This works for me. Thanks!

Issue #2301, opened 2022-01-22
Current issues with testing scripts


edomt:
[High priority for the week of January 24]

- [x] https://github.com/owid/covid-19-data/pull/2307, https://github.com/owid/covid-19-data/pull/2317
- [x] https://github.com/owid/covid-19-data/pull/2318
- [x] https://github.com/owid/covid-19-data/pull/2302
- [x] Jamaica: not updated since 2022-01-08
- [x] https://github.com/owid/covid-19-data/pull/2305
- [x] https://github.com/owid/covid-19-data/pull/2313
- [x] https://github.com/owid/covid-19-data/pull/2308
- [x] https://github.com/owid/covid-19-data/issues/2261, https://github.com/owid/covid-19-data/issues/2319
- [x] https://github.com/owid/covid-19-data/pull/2310
- [x] #2320, #2322

edomt:
El Salvador fixed in e780e67b895a4e5e2c208ea2666d5f653c53b09d

3dgiordano:
Hi @edomt 

For Peru I did something
Peru ->  https://github.com/owid/covid-19-data/pull/2302

Rouein:
Hi @edomt,

I checked Laos's source link and it is missing the test data.
I guess it has something with Laos's health ministry [website](http://www.health.gov.la/) being down .

3dgiordano:
Hi @edomt and @Rouein 
The current source of Laos is a outdated dashboard.

The 2 alternative sources
- The main site (incremental web scrapping)   https://www.covid19.gov.la/index.php
- Official Facebook account (web scrapping + OCR) https://www.facebook.com/CCEH.MoH.Lao/photos

I couldn't find any other alternative source on the internet.

Rouein:
Moldova #2305

3dgiordano:
Hi @edomt and @Rouein 

I solved how to get all the data from El Salvador.
I'll give you an example of a batch at the end of the day.
It will be an example so that you can see how to process and not have to use scrapping.
I mention it so as not to overlap each other.

Regards

Rouein:
@3dgiordano I am concentrating on Moldova for now

Rouein:
Moldova fully resolved!

3dgiordano:
El Salvador -> https://github.com/owid/covid-19-data/pull/2307

Inside the PR, I attach the CSV in case it can be useful while evaluating the code.

3dgiordano:
[...]

PR #2307, opened 2022-01-23
El Salvador Testing Data

3dgiordano:
Hi @edomt @camappel @Rouein 

This work is based on the issue reported on https://github.com/owid/covid-19-data/issues/2301

The problem with the previous method is the new need for a hidden token inside the body, needed to process the date.
Also I discovered the way how to get the valid dates and the values without scrapping, because all the data exist inside JavaScript variables and are easy to parse and convert to list values.

The new version support token extractor and valid dates for data use direct response to parse without scrapping.
Also the new version calculate the "Positive Rate".

I migrate my code to your base class, I don't test over this version.

Sure, the extraction of the token and the values is possible to do with regular expression, but I did it quickly to be able to collaborate to attend to the high priority that you marked.

Important note, take the example of the interaction mechanism they had with the dates to make a similar mechanism.
But in my tests I use multiprocessing with pool processing to speed up the times and the process takes much less time.
I leave it to your discretion to evaluate incorporating parallelization.
In the same way, migrating to regular expression if you think it is necessary.

Sorry if there are minor errors in the code, I promise in the future to install the OWID code to share something tested with your class model.

I hope to be helpful

Regards
David


3dgiordano:
I am attaching the CSV file generated by my program, in case it is useful to update the current one while evaluating the code.

[El Salvador.csv](https://github.com/owid/covid-19-data/files/7919559/El.Salvador.csv)

camappel:
Ideally this would only scrape the last 7 days in order to caluclate the PR and append those 7 days to the df, as it takes a long time at the moment.

It's great for now, though!

camappel (approved):

edomt:
My understanding is that it's meant to stop when it reaches the latest known date from the CSV?

camappel:
Yes, and it still does that. I meant "the last 7 days _of available entires_ in order to calculate the PR and append those 7 days to the df". The old script scraped from the last date in our data base until the latest available entry. I would just stick with our current script, but it doesn't scrape the positive tests, which differ from the reported by the JHU.

edomt:
Sorry, I mistyped: I meant it's supposed to _start_ at the last known date, yes.

We can store an endless number of variables in the country CSV files, so here's what should be possible:
- Make the script output the CSV that stores daily tests and raw daily cases (in a variable like `raw_cases`), append to existing data
- In that output, also calculate the positive rate as 7day_avg_raw_cases / 7day_avg_daily_test

camappel:
That was my original plan, but I thought it might crash the grapher. That should be an easy fix

3dgiordano:
Hi @camappel @edomt 

It's quite easy to do the same as the previous version which was incremental.
They get the date from the last data, they get the list of available dates from the home page (like the new way) and they cut the list from the latest date to the end list to get the new dates to add.
They then calculate as does the incremental version that adds the data from the past with the new, and recalculates the PR of all data.
The batch version makes sure that if they make changes in the past data it is reflected, but it takes time if it doesn't run in parallel.
The incremental version would not reevaluate data from the past, but it will run in less time.

lucasrodes:
As far as I understand, the data for El Salvador from this source has not been updated since 2021-04-22. If so, I would suggest we instead try to find an alternative source. I am trying to see if PAHO does something similar as they do with vaccination, but haven't found anything so far.

3dgiordano:
@lucasrodes Apologies I had not realized that they were not updating the data.
I found an alternative source that has data from September 2021 to today (keep 5 months back).
Perhaps the best idea is to generate a csv, take the data from September and merge it, and start using the new source as an incremental data source.

Tell me what you think about it, what could be the best option.

3dgiordano:
@camappel @lucasrodes I have a python that gets the data from an alternate source.
The alternative source seems to keep the data of the last 126 days so it can be used to incorporate manually or to create an incremental with it.
The alternative source has data up to January 20 2020 at the moment.
The data is also official (the same portal covid19.gob.sv/)

In a few hours I will share an example of how to process and transform the data to generate the csv on gist.

3dgiordano:
@camappel @lucasrodes 
I write a python code sample for process alternative (sorry not use the OWID class model, but it easy to traslate)

https://gist.github.com/3dgiordano/dfd3fff4dcba87d851749ab0727380ba

I don't use the OWID class model because it's up to you how you'll use the data.

It occurs to me since they have the batch data and the data of this new data model that maintains the data of the last 126 days. They can take the El Salvador data from the Batch and start using the new one as something incremental.

Attached is the file generated by the example, it can be used to update the data for El Salvador manually.

[El Salvador_alt.csv](https://github.com/owid/covid-19-data/files/7936376/El.Salvador_alt.csv)

Rouein:
is there any updates here?
If no one is working on it , I can use the source @3dgiordano provided to write the script.

3dgiordano:
@Rouein I don't know who's on the subject.
Yesterday I was analyzing the origin of the new data source that I suggest and apparently the new data source starts recording the data from September onwards. On September starts with a few records (no report backward data)
It could be used as a batch that does incremental data that was obtained from the other version.
I am waiting for a reply from @camappel or @lucasrodes
I think the new source (from the same site but another page) is a good one to increment the data onwards.

camappel:
Hi @3dgiordano ,

The new source you found looks great; thanks very much!

A note on the [El Salvador_alt.csv](https://github.com/owid/covid-19-data/files/7936376/El.Salvador_alt.csv) you shared: the figures between 2021-10-27 and 2021-11-05 only give the first digit. This is because the figures for those dates include a comma after the first digit, and so line 16 of [this script](https://gist.github.com/3dgiordano/dfd3fff4dcba87d851749ab0727380ba) splits the figure in half. 

I don't believe the government updates their data retrospectively, therefore it may be easier to write an incremental script that scrapes the daily change.

3dgiordano:
Ok, Thanks @camappel 
I'm going to work on creating the incremental version for El Salvador.

I have had no response from those who maintain the page that they will fix the other report.

Rouein:
Hi @3dgiordano ,
I used your code and refactored it to the format  : #2367

3dgiordano:
@camappel I incorporated the incremental version into the PR.
One problem I ran into is that if I don't keep the "positive" data then I can't calculate the "Positive rate" for the incremental.
I'm not using export_datafile as it filters the columns and I lose the positive column information needed to recalculate as the data increases.
I used the version implemented by @Rouein at https://github.com/owid/covid-19-data/pull/2367 but I made it incremental so we don't lose the history generated by batch version.
I keep the batch version so you can see what generated the historical data.
It should be possible to remove it since the incremental will take care of it from now on.

I don't use the "locale" mechanism to transform dates (the %b to parse months), but you can use the mechanism in @Rouein's code if you prefer if works for you.

I leave it up to you to adapt it to your needs and knowledge, both how to solve the "Positive rate" calculation, whether keeping the "positive" column in the csv can cause problems, or the use of %b for parsing the month in Spanish (which in my case does not work).

I updated also the csv file generated by the incremental version (over the csv generated by the batch version).

I hope it is useful.

camappel:
Thank you @3dgiordano and @Rouein for collaborating on this!

I've incorporated @Rouein 's pipe_date function as it's more efficient. I think that incremental is a more appropriate format in this situation.

And it's OK for there to be a 'positive' column in the exported CSV, as this column won't be picked up by the grapher.

I will merge pending @lucasrodes 's approval.

camappel (approved):

Rouein:
My last commit made the code incremental, in case you want to remove the columns.(you can just copy paste it after the last commit and test it.) 
All the credit to @3dgiordano, I just refactored his code!

lucasrodes (approved):
LGTM! Thanks @camappel @3dgiordano @Rouein for your work

PR #2308, opened 2022-01-23
fixed(test,libya)

Rouein:
I did a touch up on the script and fixed it. also collected the missing data.


Rouein:
you are welcome.

Rouein:
To make sure I can get as much data as possible I manually added some of the data posts from their facebook posts but still couldn't get a continuous data series.

lucasrodes:
> To make sure I can get as much data as possible I manually added some of the data points from their facebook posts but still couldn't get a continuous data series.

It would be great if you could edit the `Source URL` in the data file for those data points coming from Facebook posts.

Rouein:
OK Done!

lucasrodes (approved):
Thanks, @Rouein!
LGTM

---
I would suggest not merging with the master so frequently as it "pollutes" the PR thread. It is totally fine to do the merge on our end when reviewing.

camappel (approved):
Another great contribution - thanks!

PR #2309, opened 2022-01-23
(test, Singapore) python script

Rouein:
the script is now adding art and pcr together as total and records it in cumulative total 
it should be merged after #2303


edomt:
Hi @Rouein 

Thank you for the pull request - however, we need to be able to merge pull requests separately, and this one contains changes to multiple scripts (Singapore and Australia). Can you please reopen a separate pull request, with only changes for Singapore compared to the master branch? (You can do that by creating your new branch from master, rather than creating it from the previous branch where you changed the Australia script)

Rouein:
sure thing!

PR #2310, opened 2022-01-23
(test, Singapore) python script

Rouein:
the script is now adding art and pcr together as total and records it in cumulative total


camappel:
I need to see if ART tests can be used to confirm cases, or else our estimate of the PR will be half the true value. If they can, we will need to get historical ART data back, or the PR will not be accurate

camappel (approved):
ART tests can be used to confirm cases, but we don't have historical data.

We should be able to get historical data soon, however. As such, I am excluding ART tests from testing figures and removing our estimate of the positive rate until this problem is solved.

camappel:
These new changes take the daily average of PCR and ART tests per week to produce a series of Daily change since 2021-11-16

scripts/scripts/testing/automations/batch/singapore.py
```diff
@@ -0,0 +1,85 @@
...
+                    "average_daily_number_of_art_swabs_tested_over_the_past_week": df[
+                        "average_daily_number_of_art_swabs_tested_over_the_past_week"
+                    ][i],
+                    "average_daily_number_of_pcr_swabs_tested": df["average_daily_number_of_pcr_swabs_tested"][i],
+                    "Daily change in cumulative total": df["Daily change in cumulative total"][i],
+                }
+            )
+            df = df.append(new).sort_values(by="date").reset_index(drop=True)
+            i += 7
+        print(df.dropna().tail(15))
```
Rouein, line 68: you forgot print function here!

scripts/scripts/testing/automated_sheets/Singapore.csv
```diff
@@ -1,86 +1,64 @@
...
+Singapore,2022-01-08,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-09,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-10,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-11,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-12,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-13,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-14,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-15,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-16,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
+Singapore,2022-01-17,,217500.0,samples tested,https://www.moh.gov.sg/covid-19/statistics,Ministry of Health Singapore,
```
Rouein, line 64: Use `cowidev.utils.clean.numbers.metrics_to_num_int` to avoid floats

scripts/scripts/testing/automations/batch/singapore.py
camappel: Thanks!

scripts/scripts/testing/automated_sheets/Singapore.csv
camappel: Good point

lucasrodes: No need to use the method in this case, as it is automatically called when exporting the data with `self.export_datafile`

Rouein: I really should sit down and sit all the util codes!

Rouein:
is this a temporary solution ? until there is enough data?
and Why do you delete the historical data before use of art ?

camappel:
Hopefully it's temporary - it depends on the government backfilling the data.

I removed the historical data because I don't think it's a good idea to report incomplete data (excluding ART tests), because it implies that the figures include **all** tests performed in the country, when really PCR tests have constitute only a fraction of total tests (this is the case in most countries).

Rouein:
> Hopefully it's temporary - it depends on the government backfilling the data.
> 
> I removed the historical data because I don't think it's a good idea to report incomplete data (excluding ART tests), because it implies that the figures include **all** tests performed in the country, when really PCR tests constitute only a fraction of total tests (this is the case in most countries).
> 
> For reference, ART tests have been used in Singapore since [2020-11-09](africacdc.maps.arcgis.com/apps/dashboards/a5603222b29c49539df3af45bac16bcc).

the link is broken(wrong)

lucasrodes (approved):
LGTM.

Will remain vigilant to see if Singapore releases more historical data.

camappel:
Correct links verifying use of ARTs:

- [2020-11-09](https://www.moh.gov.sg/licensing-and-regulation/regulations-guidelines-and-circulars/details/list-of-covid-19-swab-providers)
- [2021-05-31](https://www.gov.sg/article/enhanced-measures-for-testing-and-contact-tracing)

PR #2311, opened 2022-01-23
(test, pakistan) python script

Rouein:


lucasrodes (approved):
LGTM.

I have added some minor changes. When the date is behind a regex, you can use the function [`extract_clean_date()`](https://github.com/owid/covid-19-data/blob/bc12e50e7bca65511f0697658b32a1345fe916a0/scripts/src/cowidev/utils/clean/dates.py#L82) instead

Rouein:
great! I was not aware of it. It would be great to make a documentation for all the util, that is of course, when you can find some free time.

camappel (approved):
Works on my end, too. Thanks @Rouein and @lucasrodes

PR #2313, opened 2022-01-24
fix(test,Laos) new script - new source (daily change)

Rouein:
the past data for daily change has been collected as well.


Rouein:
The governmental website is on, but the data has not been updated since 20th. If it doesn't update in the next two days I can modify the code to scrape the data from the news website, I used to get the historical data.

Rouein:
no need. I just checked and the website has updated ( the latest data from 24th is available)

camappel (approved):

camappel:
I think we should use this script, then I can fill in missing dates manually once a week

camappel:
Hi @Rouen,

Even though this script works perfectly, I'm going to close this PR in favour of manual updates, as our pipeline makes it complicated to combine Cumulative total and Daily change.

Thank you very much anyway - we'll probably come to this PR in the future.

Rouein:
is there a source that provides Cumulative total ?

Rouein:
I am asking because the data doesn't make sense to me: I took the Cumulative total of the old data from 2021-05-19 and then I added the sum of the daily changes I scraped from 2021-05-21 to 2022-01-07 and it resulted in (1093706) which is higher than the last Cumulative total in the old data (990520) at 2022-01-09

camappel:
Thanks for pointing this out; these discrepancies sometimes happens when countries have multiple data infrastructures. Given that the original source for the cumulative total reports different daily changes than the proposed new one and is updated less frequently, I think it makes sense to switch to daily change and not have an estimate of the cumulative total, similar to the situation we have with [Palestine](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=asc&pickerMetric=location&Metric=Tests&Interval=7-day+rolling+average&Relative+to+Population=true&Color+by+test+positivity=false&country=~PSE).

I will merge once we have approval from @lucasrodes .

camappel (approved):

lucasrodes (approved):
LGTM!

Thanks @Rouein, @camappel

PR #2314, opened 2022-01-24
New vax data

Phuclygia:
This data hasn't been reported for a week. The data was found in moh.gov.vn.


edomt:
Hi @Phuclygia 

Unfortunately, the reports made by the government have a couple of big problems:

- Their format changes very regularly, which makes it very hard to build a script that doesn't crash every few days.
- The information provided doesn't allow us to calculate several of our metrics. Vietnam uses the Abdala vaccine, which is a 3-dose vaccine. But the data reported is only broken down by 1st, 2nd, and 3rd doses. Therefore:
  - We cannot calculate `people_fully_vaccinated` because we can't know how many 2nd doses were final doses (for vaccines other than Abdala) and how many were 2nd Abdala doses.
  - We cannot calculate `total_boosters` because we can't know how many 3rd doses were final doses (for Abdala) and how many were booster doses for other vaccines.
  
Therefore we rely on the WHO's data, where `people_fully_vaccinated` is directly reported as such.
We realize that this isn't ideal, but with the current reporting format, unfortunately we have no alternative.

PR #2318, opened 2022-01-24
fix(ethiopia): change source to Africa CDC

camappel:
The script for Ethiopia stopped working on 2021-12-31, outputting the following: 'On one or more dates, `Cumulative total` is greater on date t-1.'

This pull request will change the source to the Africa CDC, which has up-to-date data


lucasrodes (approved):
LGTM.

I updated the data file with the latest data point.

PR #2322, opened 2022-01-24
test: austria

lucasrodes:
- Restore data file
- Fix script by excluding data for 2022-01-22
- Improve `make_monotonic` function for testing

xref #2320


camappel (approved):
Thanks

PR #2324, opened 2022-01-25
the update of Tawian.csv for 2022-01-24.

gugod:
It looks like this was missing in todays' bulk update.

The parser seemed OK, I re-ran cowid-vax and it finished successfully.


lucasrodes:
Thanks, @gugod.

Perhaps the data was not available at the time the update ran.

PR #2325, opened 2022-01-25
Solving Conversion to float in case of NAs

Rouein:
Same Can be applied to Vax Util to avoid type conversions such as [Indonesia](https://github.com/owid/covid-19-data/blob/master/scripts/output/vaccinations/main_data/Indonesia.csv)

### DIFF
--- a/scripts/src/cowidev/testing/utils/base.py
+++ b/scripts/src/cowidev/testing/utils/base.py
@@ -3,6 +3,7 @@ import os
 import pandas as pd
 
 from cowidev.utils import paths
+from cowidev.utils.clean.numbers import metrics_to_num_int
 
 
 COLUMNS_ORDER = [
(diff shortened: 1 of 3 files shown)


lucasrodes:
Hi @Rouein,
Thanks for your help.

I have generalized the lines you implemented and removed the condition of `NaN` being present in the Series. This way, we ensure that the Series is always "cleaned". For instance, in the case of Chile ([`chile.py`](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/automated_sheets/Chile.csv), the column `Daily change in cumulative total` has no NaN, but still, it would be nice if we could cast the column values to Int.

Furthermore, I have changed the location of the type casting to ensure that it is always applied.

Rouein:
I recommend using this in Vax Incremental as well. and perhaps changing [indonesia.py](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/vax/batch/indonesia.py)

lucasrodes:
Thanks for your suggestion.

This is definitely on our roadmap, once we have the time to add some changes to the vaccination pipeline. I'll let you know once we make progress on this front.

lucasrodes (approved):
LGTM

PR #2331, opened 2022-01-26
refactor(test, sweden): python migration

Rouein:
The data is manually collected through reviewing all the PDF files published weekly by the public health agency of Sweden. The data selected here are the most recent mentions of them in any of the published files.  
Please note that the agency, until 2020-07-10 (Data pertaining to 29 June – 5 July) published only the number of people tested, therefore `people tested` label and from thence to 2020-10-02 ( Data of 21 – 27 September) published, only the number of the tests and thereafter both.
`Cumulative total` is not mentioned in any of the publications, therefore the best estimate based on the weekly numbers is calculated.
They have also published the data on antibody tests (conducted only in the first few weeks) and the antigen tests (currently conducted). Since these data were not used in their calculation of positive rates, they are not included here.


Rouein:
wheeew! that was a bit wordy. 

I have also commented out the part that calculates the next `Cumulative total` in the script . I leave it to you to decide if you want it calculated.

edomt:
I'll let @camappel look into this further, but just a quick note that we want to avoid mixing units in the same time series (people tested / tests performed).

Rouein:
the alternative is to get the `Cumulative total` till July 5th and remove the points. It will be only `tests performed` after that.

lucasrodes (changes requested):
Hi, @Rouein,
Thanks for your help and new contribution, it is very appreciated.

Some comments, on why this PR can't be approved as it is now:

1. We currently only allow for a single unit in the file (more details [here](https://github.com/owid/covid-19-data/blob/master/scripts/docs/testing/CONTRIBUTE.md#metrics-collected)).
2. The field `Daily change in cumulative total` should contain _daily_ values, and not weekly. Our current approach in Sweden's data is to estimate the daily average based on the weekly absolute number (i.e. weekly_number/7) and assign it to all the days in the week. You can see this in the [data file for Sweden](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/automated_sheets/Sweden.csv) and on our [live charts](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&facet=none&pickerSort=asc&pickerMetric=location&Metric=Tests&Interval=New+per+day&Relative+to+Population=true&Color+by+test+positivity=false&country=~SWE.).


I would suggest dropping the changes to the CSV file and updating the script to add the data values in the correct format. For inspiration, you can check our approach for [Singapore](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/automations/batch/singapore.py) (although batch import, similar), where the data is also reported weekly (note that in the case of Singapore they report the daily average and not the weekly total as Sweden does).


Thanks for your help!

Rouein:
My understanding was that since the data is available, a `Cumulative total` would be enough. I also assumed that the `Daily change in cumulative total` is used to make a new 'Cumulative total` in the backened.

Rouein:
If that is not the case, then we can simply take the 7-day average of each data point and populate the days of the week.

Rouein:
Also I am not a Swedish speaker , I did my best reading through the documents with a translator tool. So I would appreciate it, if you could verify the integrity of the data,

lucasrodes:
@Rouein,
Thanks for getting back.

> If that is not the case, then we can simply take the 7-day average of each data point and populate the days of the week.

As I already mentioned, the current file for test data for Sweden already collects daily data by averaging total weekly values. You can see this in the [data file for Sweden](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/testing/automated_sheets/Sweden.csv) and on our [live charts](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&facet=none&pickerSort=asc&pickerMetric=location&Metric=Tests&Interval=New+per+day&Relative+to+Population=true&Color+by+test+positivity=false&country=~SWE.).

> My understanding was that since the data is available, a `Cumulative total` would be enough. I also assumed that the `Daily change in cumulative total` is used to make a new `Cumulative total` in the backened. 

Estimating the `Cumulative total` will be done later in the pipeline, automatically. So there is no need to add the column in the country file. We are in case 2 from the cases listed [here](https://github.com/owid/covid-19-data/blob/master/scripts/docs/testing/CONTRIBUTE.md#cumulative-total-vs-daily-change-in-cumulative-total).

---
All in all, I would suggest _dropping_ the changes to the CSV (as the current ones are already fine) and focusing only on the script (I will run some tests on the script later).

camappel:
I agree with @lucasrodes that we should leave the CSV in its current state.

lucasrodes (approved):
Working on my end.

To test it, you can first do one of:
- Artificially modify values for `Daily change in cumulative total` for last week.
- Temporally remove a couple of last rows

Then run the script for Sweden

Rouein:
I just pushed my version!

scripts/scripts/testing/automations/incremental/sweden.py
```diff
@@ -67,23 +63,37 @@ def _get_relevant_element(self, soup: BeautifulSoup) -> element.Tag:
...
-        count = re.search(fr"[vV]ecka{self.week_num} \d+ (\d+)", text).group(1)
+        count = re.search(fr"[vV]ecka{week_num} \d+ (\d+)", text).group(1)
         return clean_count(count)
 
     def _parse_date(self) -> Iterator:
         """parses the date from the week number."""
         week = Week(2022, self.week_num, system="iso").iterdates()
         return week
 
+    def _load_last_date(self):
```
Rouein, line 92: I think this is unnecessary.

lucasrodes:
Hi @Rouein,
I don't see the value-added after https://github.com/owid/covid-19-data/pull/2331/commits/08f5238b7dd0b0e100aa99493a566b66901427a4. Could you please explain the motivation behind the changes introduced after my approval?

I can see that you have:
1. Edited the data file: Removal of old data points, change of URLs, etc.
2. Changed how the daily values are estimated. Using a `for` loop instead of `pandas`, which provides more efficient tools (C under the hood). Plus removed some dates checks.

I will roll back unless 

For instance, the previous version avoided the usage of `for` loops, in favour of `pandas.bfill` function which is more efficient. Also, the data in the CSV was more complete. I don't understand the motivation behind the changes introduced after my approval.

Rouein:
I did not see your version. and I had commit before and pushed later. right now it is a bit messy. Let me clean it up and get back to you.

Rouein:
> 
> I can see that you have:
> 
> 1. Edited the data file: Removal of old data points, change of URLs, etc.
> 2. Changed how the daily values are estimated. Using a `for` loop instead of `pandas`, which provides more efficient tools (C under the hood) and removed some date checks.
> 
> If there is no substantial improvement, I will lean towards rolling back to the previous version.

OK, it all cleaned up. Your Version looks fine. I just did a bit of visual touch up.

I changed the data though, the previous data contains numbers of people tested and has them labeled as tests performed. Again, I'm just trying to be the most accurate I can be. All the data I have provided are verifiable via links.
the old data points  referenced from web-pages were later on corrected in the PDF files.

As for the performance, When I'm free I will use psutil to measure them against each other but my guess is that the difference is negligible. 

I apologize for any mess ups like this. I don't usually work with pandas or scraping data. It is more of a hubby for me. 
When I get tired of and frustrated at C++ I write some python to feel better(If that makes sense)

lucasrodes:
Hi @Rouein,
Thanks for reviewing this, we really appreciate it

> I changed the data though, the previous data contains numbers of people tested and has them labeled as tests performed. Again, I'm just trying to be the most accurate I can be. All the data I have provided are verifiable via links.
the old data points referenced from web-pages were later on corrected in the PDF files.

Thanks for catching this, very valid and important point. Thanks for insisting.

> As for the performance, When I'm free I will use psutil to measure them against each other but my guess is that the difference is negligible.

Agree, the difference will probably be negligible.

> I apologize for any mess ups like this. I don't usually work with pandas or scraping data. It is more of a hubby for me.
When I get tired of and frustrated at C++ I write some python to feel better(If that makes sense)

No need to apologize, but thanks. I did not mean to make you feel bad, so I apologize if I did. Also, totally understand playing with python as a hobby :)


---

I'll review and merge

lucasrodes (approved):
LGTM as it is now. 

Please @camappel let me know if it works on your end

Rouein:
P.S. I think it would be a good idea to get the positive rates from the source as well.

camappel:
This works on my end. Thanks a lot @Rouein for this and going through the old PDFs - I know this can be tedious work.

PR #2334, opened 2022-01-26
data(tests,usa): calculate positive rate

camappel:
Change numerator from JHU cases to CDC positive tests


edomt:
Has this been reviewed before merge?
I would have added at least a `.sort_values("Date")` on both `pos` and `df` to make sure the rolling sums are calculated in the right order.

camappel:
I didn't think it was a large enough change to warrant a review. I'll request a review for every PR from now on

PR #2335, opened 2022-01-26
data(tests,usa): calculate positive rate 

camappel:
Improvement on #2334


edomt (changes requested):
Thanks - is there a reason behind the `.fillna("")` on line 54?
Also there seems to be an issue with the general structure: some parts are functions and some other parts are methods inside `UnitedStates` - @lucasrodes can help with sorting this out if needed.

camappel:
Thanks for pointing this out. I was using the structure that was already implemented, but I'm happy to fix it

lucasrodes (approved):
LGTM!

Added some minor styling changes.

edomt (approved):

PR #2339, opened 2022-01-26
fix(tests,australia): remove daily change

camappel:
Remove daily change column and drop days where cumulative total = 0


lucasrodes (approved):
LGTM

PR #2343, opened 2022-01-26
refactor(test, hong kong): python migration

lucasrodes:
Migrate R to Python


camappel:
From 2022-01-01, figures for positive tests is reported in the 'Number of cases tested positive for SARS-CoV-2 virus' columns instead of the 'Number of confirmed cases' column in the CSV. I'll try and fix

PR #2346, opened 2022-01-27
Date vax update

Phuclygia:
This data is 24-01-2022, not 25-01-2022


edomt:
Thanks @Phuclygia !
For future updates like these, can you please edit [this file](https://github.com/owid/covid-19-data/blob/master/scripts/output/vaccinations/main_data/Vietnam.csv), rather than the output file (otherwise your edit will get overwritten at the next update). Thanks!

PR #2348, opened 2022-01-27
refactor(test, Cape Verde): python migration

Rouein:


camappel (approved):
Great as always. Thanks @Rouein!

lucasrodes (approved):
LGTM. 

Cleaned and change some regex, plus using `extract_clean_date`.

Might need to enable Portuguese locale on your system (linux: step 1 and 2 in [this guideline](https://www.behova.net/configure-locale-on-arch-linux/)).


Thanks for you work!

Rouein:
> LGTM.
> 
> Cleaned and change some regex, plus using `extract_clean_date`.
> 
> Might need to enable Portuguese locale on your system (linux: step 1 and 2 in [this guideline](https://www.behova.net/configure-locale-on-arch-linux/)).
> 
> Thanks for you work!

I was trying to get locale rather than making the dict, nice to know it works.
Since I'm working with ROS I don't touch locale on my system, unfortunately, some modules are very sensitive.

PR #2353, opened 2022-01-28
refactor(test, Fiji): migrate to python

Rouein:
This script increments unlike the R script(batch), due to the source. 
I scraped Historical data but to fill the gaps and `Cumulative total` manual work is needed. I live that part to whichever resigned soul that has time.


Rouein:
I pretend this did not happen

PR #2358, opened 2022-01-28
fix(test, Russia)

Rouein:
Name correction + removing the space between digits


camappel (approved):
Nice fix. Thanks @Rouein

Rouein:
You are welcome!

PR #2367, opened 2022-01-29
refactor(test, El Salvador): new source 

Rouein:


3dgiordano:
Hi @Rouein 
I'm going to take this version as the basis for the incremental I'm proposing on the PR that I'm working on https://github.com/owid/covid-19-data/pull/2307

@camappel @edomt Do not take this version to Merge, because as was discussed in the PR https://github.com/owid/covid-19-data/pull/2307, the strategy is changed towards an incremental one, taking the data that the new batch version proposed in the PR (probably delete the batch after that), and adding the new data from a new incremental version (because the data is from September to today in the new source).
I'm going to take @Rouein's refactored code and migrate it to the incremental version I'm about to send you guys for review in that PR.

Thank you @Rouein

camappel:
Thank you @3dgiordano and @Rouein for collaborating on this! 

I've incorporated @Rouein 's pipe_date function as it's more efficient. I think that incremental is a more appropriate format in this situation.

And it's OK for there to be a 'positive' column in the exported CSV, as this column won't be picked up by the grapher.

camappel (approved):

camappel:
Thanks a lot for this @Rouein, but we're going to go with the incremental version. Your help is greatly appreciated nonetheless.

PR #2369, opened 2022-01-29
Data vax update

Phuclygia:
Another data for the vaccine from 28-01-2022.


lucasrodes (commented):
@edomt ~Note that these PRs are changing the read-only files under `public` folder.~ Just read your comment https://github.com/owid/covid-19-data/pull/2374#issuecomment-1025509411 😄 

@Phuclygia First, thanks for your contribution! It would be great if the data changes could be added to the file we use to update the dataset instead: Find it at https://github.com/owid/covid-19-data/blob/master/scripts/output/vaccinations/main_data/Vietnam.csv 😄

PR #2372, opened 2022-01-30
refactor(vax, Azerbaijan): Changing dependencies

Rouein:
Refactoring the code to use `pdftotext`.
This enables us to remove unnecessary dependencies (currently 4 for PDFs)


lucasrodes:
Thanks for your help! I tweaked the regex a bit, as it was not working for today's report. Hopefully, it will continue to work in the upcoming days!

Rouein:
I have made a list and will continue refactoring so that they only use `pdftotext` . For your reference: 

- [x] Azerbaijan vax
- [ ] Kenya vax
- [ ] Macao vax
- [ ] Mexico vax
- [ ] Sri lanka vax
- [ ] Taiwan vax

PR #2378, opened 2022-01-31
Automation(vax, Vietnam):New Script

Rouein:


lucasrodes (approved):
LGTM. Added missing metrics, and minor tweaks on regex.

Rouein:
you should roll back the last commit. Abdala is three dose vaccine!

PR #2382, opened 2022-01-31
(tests, fiji): collect cumulative total

camappel:
Collect cumulative total instead of daily change.

Data backfill required before merge


Rouein:
I am glad they changed their format!

lucasrodes (approved):
LGTM! Thanks for reaching out to them @camappel!

PR #2383, opened 2022-01-31
refactor(vax, Kenya): Changing dependencies

Rouein:
Now using `pdftotext`.

### DIFF
--- a/scripts/src/cowidev/vax/incremental/kenya.py
+++ b/scripts/src/cowidev/vax/incremental/kenya.py
@@ -1,25 +1,31 @@
 import os
 import tempfile
 import re
-import urllib3
 
 import pandas as pd
 import pdftotext
 
-from cowidev.utils import clean_count, get_soup
-from cowidev.vax.utils.incremental import merge_with_current_data
+from cowidev.utils import clean_count, get_soup, paths
+from cowidev.utils.clean import extract_clean_date
 from cowidev.utils.web.download import download_file_from_url
-from cowidev.utils import paths
-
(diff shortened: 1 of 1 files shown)


lucasrodes (approved):
Minor changes so the regex can work.

Note that there are some minor differences in the string format depending on whether you use pdftotext or PyPDF2. Therefore, we need to either (i) adapt the regex to the new format or (ii) change how we map the PDF content to text.

Testing the script: `cowid-vax get -c kenya` should raise an error (in the case of Kenya, as the script is incremental, you would need to remove the last datapoint so it actually runs)

Rouein:
> Thanks!
> 
> Minor changes so the regex can work.
> 
> Note that there are some minor differences in the string format depending on whether you use pdftotext or PyPDF2. Therefore, we need to either (i) adapt the regex to the new format or (ii) change how we map the PDF content to text. I have done the former, option (ii)
> 
> Testing the script: `cowid-vax get -c kenya` should raise an error (in the case of Kenya, as the script is incremental, you would need to remove the last datapoint so it actually runs)

I do a bit of change and then test it with `python ***.py`. I try to leave the original code untouched, if it produces the same result, I push it.

Rouein:
Actually the `if __main__` takes care of that, I have the library installed just couldn't bother with all the tokens, so I opted to use python to run the scripts and so far I have produced the data for all of the scripts I have pushed.

lucasrodes:
That is because the testing pipeline works different. There we do have `if __name__ ...` statement.

If you are not using cowid-vax, then you can add the `if __name__...` locally and test.

PR #2393, opened 2022-02-02
Feature/tests refactor

Pakinay:
Tap on a clip to paste it in the text box.


Pakinay:
Use the edit icon to pin, add or delete clips.

PR #2397, opened 2022-02-03
(tests, suriname): automate collection

camappel:
Still a draft, as it requires checks on date to ensure the same day isn't scraped twice


lucasrodes (approved):
LGTM. Working on my end, I generalized some statements.

PR #2379, opened 2022-01-31
Automation(test, Vietnam):New Script

Rouein:

PR #2398, opened 2022-02-04
Data vax update

Phuclygia:
Another data for the vaccine from 02-02-2022.


Rouein:
Hi @Phuclygia,
Thanks for your help fellow contributor, but there is a script that already collects the latest vaccination data. perhaps you could help with the efforts at [this](https://github.com/owid/covid-19-data/pull/2379) front.

PR #2411, opened 2022-02-07
(tests, zambia): drop erroneous data point

camappel:
Incorrect data point at 2021-10-30 was causing the script to crash


lucasrodes (approved):
LGTM

PR #2413, opened 2022-02-07
fix(tests, costa rica)

camappel:
The source_url function on line 18 needs to be amended to use the most recent Friday as the date


lucasrodes (approved):
LGTM, thanks @camappel

PR #2414, opened 2022-02-07
fix(tests, cambodia)

camappel:
Correct soup.select


lucasrodes (approved):
LGTM! Thanks @camappel

PR #2416, opened 2022-02-09
fix taiwan.py: update parse_date according to change of file name.

gugod:
currently parse_date is derived from file name, which used to contain
a whitespace character as the delimiter between date and
non-date. Such as this:

    1110207 COVID-19疫苗接種統計資料.pdf

But the file for 2022/02/08 is named differently like this:

    1110208COVID-19疫苗接種統計資料.pdf

Since this change might persists or not, the regex is augmented to
accept both variants.


edomt:
Thanks!

PR #2418, opened 2022-02-10
Ukraine add total_boosters vaccination info

Golova1111:
This is a PR that add the **total_boosters** data for Ukraine, however, IDK exactly how should we properly handle this data

From the point of Ukraine's Ministry of Health, it is possible to get an **additional dose** and a **booster dose** (this is not a new concept, afaik this is the same as in the USA)

An additional dose is a dose that is dedicated for people with low immunity reaction (most often this is people with HIV/cancer). The additional dose can be done after 28 days after the second dose. Ukraine statistics count this type of vaccination most of all as a separate entity or rarely sum it up with a second dose. As for now, around **20k** people have this type of vaccination 

Government dashboard with additional and booster doses on it (https://vaccination.covid19.gov.ua/)
Additional info: https://uatv.ua/en/health-ministry-approved-a-booster-dose-uatv-studied-difference-with-an-additional-vaccine/

A booster dose is a "classic" booster vaccination that is available for all people in the country, around **500k** people have already done the booster in Ukraine.

Current realization sums up the additional and booster doses together and counts them as a booster. Summing up additional and second may lead to misunderstanding in reading the statistics, however, if the database standard is to count these doses as second doses or not count at all, I can easily fix this in the scope of this PR

### DIFF
--- a/scripts/src/cowidev/vax/batch/ukraine.py
+++ b/scripts/src/cowidev/vax/batch/ukraine.py
@@ -68,15 +68,10 @@ class Ukraine:
         total_df = main_df.join(booster_df.set_index("date"), on=["date"])
 
         for col in ["total", "moderna", "astrazeneca", "pfizer", "jnj", "sinovac"]:
-            total_df[f"booster_dose_{col}"] = (
-                total_df[f"third_dose_{col}"] +
-                total_df[f"additional_dose_{col}"]
-            )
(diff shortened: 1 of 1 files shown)


lucasrodes (approved):
Hi @Golova1111,
Thanks for your contribution.

It LGTM!

We currently consider as "booster dose" all doses beyond the protocol that was defined initially. We apply this logic to _all_ countries.

PR #2425, opened 2022-02-11
(tests, netherlands): calculate PR directly

camappel:
According to the [RIVM](https://data.rivm.nl/covid-19/):

"From 8 February 2022, the positive SARS-CoV-2 test results will be directly reported from CoronIT to the RIVM. The test results of other test providers (such as Testen voor Toegang) and healthcare institutions (such as hospitals, nursing homes and general practitioners) that provide their positive SARS-CoV-2 test results via the Meldportaal of GGD GHOR are also reported directly to the RIVM."

As such, we now calculate the Positive Rate directly based on the number of positive tests reported by the RIVM


lucasrodes (approved):
LGTM, thanks @camappel

Issue #2409, opened 2022-02-07
refactor(test): simplify & fix scripts


lucasrodes:
## The problem
The testing pipeline is currently running very slow, mainly due to two problems:

1. Some sources are not working anymore or don't work properly (slow loading times, large data files, etc.)
2. Some country scripts have very complex code or code that differs from standard library style. This may slow down the execution and difficult maintenance.

All of this makes the process slower and more difficult to debug.

## Goal
The goal of this issue is to provide a list of these scripts so maintainers and contributors can help improve these scripts by either re-writing them or changing the sources in use.

## List of problematic scripts
Find a list of the identified scripts that need refactoring (sorted by priority*):

- Source not working (problem 1):
  - [x] [`kazakhstan.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/batch/kazakhstan.py):  source not working (timeout), script's code is very complex.
  - [x] [`zambia.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/batch/zambia.py): [`make_monotonic()`](https://github.com/owid/covid-19-data/blob/ced7d97a064860d2ac6309ae10bf2b7462a9f5b9/scripts/src/cowidev/testing/utils/utils.py#L4)'s `max_removed_rows` limit exceeded. #2411 
  - [x] [`costa_rica.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/batch/costa_rica.py): 404 error. #2413 
  - [x] [`cambodia.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/incremental/cambodia.py): data extraction is not robust #2414 

- Source very slow (problem 1):
  - [x] [`peru.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/batch/peru.py): source is very slow (large file). #2428
  - [x] [`moldova.py`](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/testing/incremental/moldova.py): very slow script (probably the source). #2427

- Non-standard code (problem 2):
[...]

PR #2432, opened 2022-02-12
refactor(test, Qatar): source change

Rouein:
#2409


scripts/output/testing/main_data/Qatar.csv
```diff
@@ -1,589 +1,611 @@
...
+Qatar,2022-01-17,3270498,1.021,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-18,3273734,1.055,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-19,3276743,1.089,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-20,3279242,1.12,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-21,3281825,1.168,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-22,3284276,1.233,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-23,3287344,1.17,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-24,3290271,1.092,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
+Qatar,2022-01-25,3292852,1.05,people tested,https://covid19.moph.gov.qa/EN/Pages/default.aspx,Qatar Ministry of Public Health,
```
Rouein: I was baffled by the positive rates so I manually calculated it and, YEAH number of positive case was larger than the number of people tested!

Rouein:
The source is the official dashboard but I am not sure how reliable the data is!

Rouein:
@lucasrodes ,
I am not sure about the priorities, but the last test data point from Qatar is from 2021-10-27. 
It would be great if you could tell me which of the 3 metrics I should use.

Rouein:
The underlying problem still remains, the metric is not suitable.

camappel:
Agreed, I was just testing something.

I think that `NUMBER OF PEOPLE TESTED IN LAST 24 HOURS` is our best choice, because as you say, the other two metrics yield inaccurate positive rates. I agree that these figures refer to tests performed instead of people tested; this is corroborated by the column name, `TotalTests24H`.

It is unfortunate that data before `2021-03-23` is missing, but it is better to have a short accurate time series than a long inaccurate one.

`NumPositiveCases24H` is equal to the number of cases reported by our general source, the JHU, so there's no need to calculate the PR directly in this case.

Rouein (commented):
LGTM

camappel (approved):
@lucasrodes TL;DR:

The old metric referred to the cumulative number of _people tested_ **at least once** since the beginning of the pandemic. This meant any person tested more than once was not counted. As such, the number of positive cases was often larger than the number of people tested.

We have therefore changed to the daily change in cumulative total of _tests performed_, regardless of how many times the tested person has been tested before.

However, this means that we have lost all data prior to 2021-03-23.

lucasrodes (approved):
LGTM!
@Rouein @camappel Thanks for the analysis & implementation

camappel (approved):
Works for me. Thanks for adding the positive rate

PR #2433, opened 2022-02-13
refactor (vax, Philippines)

Rouein:


Rouein:
Hi @lucasrodes, 
Can you elaborate?

The [source](https://e.infogram.com/_/yFVE69R1WlSdqY3aCsBF?src=embed) seems to include `people_vaccinated`, `people_fully_vaccinated` and `total_boosters`

lucasrodes:
1. We are already reporting `people_fully_vaccinated`. Therefore, I think that the title of your PR might be wrong. You might have wanted to add `people_vaccinated`, which we are not reporting (correct me if I am wrong)
2. As I explained in https://github.com/owid/covid-19-data/pull/2433#issuecomment-1038209789, the source field "first dose" does not include people that received the single shot. Therefore, we cannot estimate the number of people that have received at least one dose. Taking the value from "first dose" would be an underestimate.

Rouein:
I double checked and realized the data is updated as well. if you see no benefits in the updated structure, we can close the PR. I leave it up to you.

lucasrodes (approved):
LGTM! Seems to be working. Thanks @Rouein

PR #2435, opened 2022-02-13
Add(vax,Laos): New country

Rouein:
Please note that the 2nd dose [here](https://www.covid19.gov.la/index.php) refers to fully vaccinated(ຮັບວັກຊິນຄົບ) as it can be observed [here](https://www.facebook.com/photo/?fbid=322396643256604&set=a.228059056023697)


lucasrodes:
Thanks, @Rouein!

PR #2444, opened 2022-02-14
refactor(test,rwa): convert to batch

Rouein:
#2409


Rouein:
you can apply `clean_count` as percussion but it will nag about it being `float`. You can either apply `number.metric_to_num_ints` first and then `clean_count` or modify `clean_count` to accept `float`.

lucasrodes (approved):
LGTM. Will merge following @camappel's approval

camappel (approved):
LGTM

Rouein:
I lean towards hard coding the 2 data points.

Rouein:
then we can do what we did with peru and put a start date.

camappel:
Good idea

camappel (approved):

Rouein:
I would recommend 2021-12-01 data seems continuous after that.

PR #2458, opened 2022-02-15
refactor(test,SVN) : Ag added to test and positive definition

Rouein:
#2409


lucasrodes (approved):
LGTM, Thanks @Rouein

camappel:
Slovenia has changed their case definition a few times; I'll try to summarise.

Previously PCR and RAT tests were performed and reported in Slovenia. However, only positive PCR tests could be used to confirm cases of COVID-19 (see screenshot below). As such, we only included PCR tests in the positive rate calculation.

Screenshot 2022-01-28 at 17 26 34

According to the [source](https://covid-19.sledilnik.org/en/stats), "from 1 February 2022, a positive result on PCR or RAT test is considered a confirmed case. Persons who are additionally tested with the PCR test after a positive RAT test are not included in the statistics of positive RAT." As such, we should change the positive rate calculation to include antigen tests, as you have done.

camappel:
This new script only includes PCR until 2022-02-01 and then both PCR and antigen. This leads to a sharp 10-fold increase in daily testing figures on the given date; this is necessary, however, in order to match the official case criteria.

camappel (approved):

camappel (approved):

PR #2472, opened 2022-02-16
Add(test,armenia): New country

Rouein:
Batch data from: [NCDC Armenia — National Center for Disease Control](https://ncdc.am/)
In the official reports (translated via google translate): "tests performed" is used thus the unit.


lucasrodes:
Thanks, @Rouein.

@camappel, currently we track `Cumulative total` manually. Do you think having daily counts instead would be OK?

If so, we need to update `Automated` to `True` and stop manual collection. Also, the script works for me!

Rouein:
> Thanks, @Rouein.
> 
> @camappel, currently we track `Cumulative total` manually. Do you think having daily counts instead would be OK?
> 
> If so, we need to update `Automated` to `True` and stop manual collection. Also, the script works for me!

I couldn't find any file containing the data manually collected, I was under the impression it is not collected at all.

camappel:
I have been collecting the cumulative total of daily tests from daily reports posted [here](https://www.moh.am/#3/News) by the Armenian MoH. The data has been visible in our [explorer](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=asc&pickerMetric=location&Metric=Tests&Interval=7-day+rolling+average&Relative+to+Population=true&Color+by+test+positivity=false&country=~ARM).

coronavirus-data-explorer (6)

This script will help save time, so thanks.

The sum of the daily change matches the cumulative total, so these figures are interchangeable.

camappel (approved):

PR #2480, opened 2022-02-18
fix(tests,iceland)

camappel:
The script for Iceland testing data no longer works due to an change in the structure of the [source](https://www.covid.is/data).

This script works but is not in a simplified format, and does not collect positive tests.


lucasrodes:
LGTM. Thanks, @camappel

PR #2482, opened 2022-02-19
enhance(vax,Library)

Rouein:
incremental option for manufacturer data added


lucasrodes:
Thanks, @Rouein,
I have slightly modified the function, so it allows for different behaviours depending on the data being exported (main, age, manufacturer).

PR #2491, opened 2022-02-21
enhance(test, Myanmar): New country automation 

Rouein:


camappel:
Thanks @Rouein , this will save us a lot of time

PR #2504, opened 2022-02-24
enhance(vax, singapore): Removing filter

Rouein:
The filter is redundant since `make_monotonic` will remove the anomalies in the data. Having the filter, disables the script to record the data points, if they are corrected by the source.


lucasrodes:
Thanks, @Rouein

Issue #2430, opened 2022-02-12
Many missing datapoints on boosters


fibke:
Would be good to double check against national health authority databases to see whether confirm whether booster data is indeed NA as reported in OWID. It appears many countries have started boosting.

Currently OWID reports NA for 85 countries

   income country_code country                              
 1 HIC    ATG          Antigua and Barbuda     
 2 HIC    BRB          Barbados                
 3 HIC    NRU          Nauru                   
 4 HIC    SYC          Seychelles              
 5 LIC    AFG          Afghanistan             
 6 LIC    BFA          Burkina Faso            
 7 LIC    BDI          Burundi                 
 8 LIC    CAF          Central African Republic
 9 LIC    TCD          Chad                    
10 LIC    COD          Congo (Kinshasa)        
11 LIC    ERI          Eritrea                 
12 LIC    ETH          Ethiopia                
13 LIC    GMB          Gambia                  
14 LIC    GIN          Guinea                  
15 LIC    GNB          Guinea-Bissau           
16 LIC    PRK          Korea, North            
17 LIC    LBR          Liberia                 
18 LIC    MDG          Madagascar              
19 LIC    MWI          Malawi                  
20 LIC    MLI          Mali                    
21 LIC    MOZ          Mozambique              
22 LIC    NER          Niger                   
23 LIC    SLE          Sierra Leone            
24 LIC    SOM          Somalia                 
25 LIC    SSD          South Sudan             
26 LIC    SDN          Sudan                   
27 LIC    SYR          Syria                   
28 LIC    TGO          Togo                    
29 LIC    UGA          Uganda                  
30 LIC    YEM          Yemen                   
31 LMIC   AGO          Angola                  
32 LMIC   BLZ          Belize                  
33 LMIC   BEN          Benin                   
34 LMIC   BTN          Bhutan                  
35 LMIC   MMR          Burma                   
36 LMIC   CPV          Cabo Verde              
37 LMIC   CMR          Cameroon                
38 LMIC   COM          Comoros                 
39 LMIC   COG          Congo (Brazzaville)     
40 LMIC   CIV          Cote d'Ivoire           
41 LMIC   DJI          Djibouti                
42 LMIC   EGY          Egypt                   
43 LMIC   SWZ          Eswatini                
44 LMIC   GHA          Ghana                   
45 LMIC   HTI          Haiti                   
46 LMIC   KIR          Kiribati                
47 LMIC   KGZ          Kyrgyzstan              
[...]

PR #2513, opened 2022-02-27
enhance(vax,fiji): add boosters and correct vaccine timeline

Rouein:
#2430


Rouein:
`spc.py` should be modified to not overwrite the boosters data as well as outputting the correct vaccine timeline.

lucasrodes:
Hi @Rouein,
Thanks for reporting this.

I am trying to reconcile SPC data with official Fiji data. As per SPC, `first doses + second doses = total doses`, which might indicate that they are not considering boosters and therefore our metric `total_vaccinations` might not be accurate at the moment.

I reached out to SPC, letting them know about this case and asked if they plan to have booster data anytime soon.

---
On another note, we may need to think about how to use two scripts (batch `spc.py` and incremental `fiji.py`) to generate the data for Fiji. An option could be to:
- Have a  `main` function in `fiji.py` that doesn't do anything.
- Call methods from `cowidev.vax.incremental.fiji` in `cowidev.vax.batch.spc`, then join SPC data with Fiji data.

lucasrodes:
Hi @Rouein,
Thanks for reporting this.

I am trying to reconcile SPC data with official Fiji data. As per SPC, `first doses + second doses = total doses`, which might indicate that they are not considering boosters and therefore our metric `total_vaccinations` might not be accurate at the moment.

I reached out to SPC, letting them know about this case and asked if they plan to have booster data anytime soon.

---
On another note, we may need to think about how to use two scripts (batch `spc.py` and incremental `fiji.py`) to generate the data for Fiji. An option could be to:
- Have a  `main` function in `fiji.py` that doesn't do anything.
- Call methods from `cowidev.vax.incremental.fiji` in `cowidev.vax.batch.spc`, then join SPC data with Fiji data.

Rouein:
The issue is that the dates might not align. Maybe use merge data and drop duplicates by date in spc?

Rouein:
This is my take on it(at least until they fix the total_vaccinations).

lucasrodes (approved):
LGTM, thanks @Rouein! Let's see if I get a reply by SPC anytime soon.

PR #2518, opened 2022-02-28
Update Armenia.csv

menoua:
Update recent WHO numbers with official Ministry of Health numbers that also include booster statistics.


lucasrodes (approved):
Thanks for your help @menoua

Issue #2534, opened 2022-03-07
bug(vax,azerbaijan): current script crashes due to regex issue


edomt:
### Description

The current script crashes due to a regex issue. The data was last updated on Feb 17.

```
2022-03-07 10:20:38,206 ERROR    VAX - cowidev.vax.incremental.azerbaijan: ❌ 'NoneType' object has no attribute 'group'
Traceback (most recent call last):
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/cmd/commons/get.py", line 39, in run
    module.main()
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/vax/incremental/azerbaijan.py", line 109, in main
    Azerbaijan().export()
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/vax/incremental/azerbaijan.py", line 95, in export
    data = self.read().pipe(self.pipeline)
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/vax/incremental/azerbaijan.py", line 26, in read
    data = self._parse_data(soup)
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/vax/incremental/azerbaijan.py", line 41, in _parse_data
    total_vaccinations, people_vaccinated, people_fully_vaccinated, total_boosters = self._parse_metrics(text)
  File "/Users/edouard/git/covid-19-data/scripts/src/cowidev/vax/incremental/azerbaijan.py", line 74, in _parse_metrics
    total_vaccinations = re.search(self.regex["total"], text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

Rouein:
this is due to a change of format in the document , they have added "Pozitiv nəticədən sonra
"Buster" doza vaksinlərin sayı" , which roughly translates to "booster doses after testing positive". I am not sure how to categorize it , would it fit into fully vaccinated ?

lucasrodes:
Fixed in https://github.com/owid/covid-19-data/pull/2541

PR #2537, opened 2022-03-07
fixes (#2534)

Rouein:


Rouein:
I am not sure about the last metric but I'll try to find the definition in Azerbaijan and adjust accordingly

Rouein:
According to [this article](https://nk.gov.az/az/article/1921/), a certificate of immunity is issued after recovering from the virus, which is on par with the vaccination certificate but expires after 6 months. After which, if a single dose of vaccine(from 2 dose protocol vaccines) is administered within 56 days, the certificate will be valid indefinitely.

edomt:
@lucasrodes I'm not sure how to interpret this in the light of our usual rules about vaccination protocols. If needed, we may need to use the WHO data to make sure everything is counted properly.

lucasrodes:
Based on what we have been doing thus far, we do not count COVID + one dose as "fully vaccinated", but "vaccinated".

However, I suspect some of these doses might also be part of the booster protocol at this stage. I reached out to them via mail, hopefully, I can get an answer soon.

I've checked data from the WHO, and it is pretty old and not consistent with our current reporting (I will revisit it). 

Another option could be to disregard these doses in our reporting completely.

Rouein:
I doubt that any of it, is a booster protocol dose since that metric is already given. Based on the official document that I linked, this is a new metric which they regard as on par with fully vaccinated.
The options are:
- go with their definition (fully vaccinated)
- considering it is just a single dose of vaccine add it to people vaccinated
- disregard it completely
I personally lean towards the last option (at least until WHO publishes a new report and we can see how they categorized it.)

PR #2564, opened 2022-03-15
enhance(test): Togo automation

Rouein:


camappel:
The cumulative total given in the dashboard does not equal the sum of daily changes since 2020-03-04 in the time series provided directly by the MoH (dashboard =  709830; direct figure = 708404). This has always been a discrepancy, presumably due to human error in data entry as well as retrospective revisions. As such, I will note in the description that the cumulative figure at the start of the time series (1428 tests on 2020-03-04) is not necessarily accurate. This discrepancy does not affect the positive rate

lucasrodes (approved):
LGTM

PR #2569, opened 2022-03-16
enhance(test): Maldives automation 

Rouein:


camappel:
The Maldives MoH declared that COVID-19 is no longer a Public Health Emergency on 2022-03-14, so it looks like they will no longer update COVID statistics

camappel (approved):

PR #2570, opened 2022-03-17
enhance(test): ECDC automation

Rouein:


lucasrodes (approved):
LGTM. Over to @camappel.

Data for Poland will be less frequently updated, but it will be automated!

camappel:
This is excellent, thank you @Rouein .

I imagine the ECDC will continue report testing data after national governments stop, so we can switch to this source as this happens.

camappel (approved):

PR #2572, opened 2022-03-19
fix(test, JOR): new payload and query

Rouein:


lucasrodes (approved):
LGTM! Over to @camappel

Rouein:
I'm not sure if it qualifies as a refactor. It seems Jordan has moved to weekly reports rather than daily reports and, using a different database(hence the new query and payload) and, because of that the previous script was returning errors.

camappel (approved):

PR #2576, opened 2022-03-26
Update China.csv

aywi:
Date mismatch for China Update on 2022-03-24. See: https://github.com/owid/covid-19-data/issues/230#issuecomment-1079016292


lucasrodes:
Hi @aywi,
Thanks for your fix!

PR #2585, opened 2022-04-06
data(vax,chn): Minor correction for 2022-03-31 and update for 2022-04-05

aywi:
Minor correction for 2022-03-31 (see https://github.com/owid/covid-19-data/issues/230#issuecomment-1085943764), and update for 2022-04-05 (see https://github.com/owid/covid-19-data/issues/230#issuecomment-1090205700).


lucasrodes (approved):
Thanks!

PR #2596, opened 2022-04-13
feat(vax,chn): Much more robust regex and handling of Chinese numerals

aywi:
See <https://github.com/owid/covid-19-data/issues/230#issuecomment-1098361939>.

The current version works for the data starting from 2021-08-12 (except 2021-08-26 and 2021-11-19).


lucasrodes (approved):
Thanks, @aywi!

Issue #2618, opened 2022-04-27
data(vax,can): COVID-19 Vaccines Currently Authorized by Health Canada


aywi:
### Country

Canada

### Domain

Vaccinations

### Which data is inaccurate or missing?

According to <https://covid-vaccine.canada.ca/> and <https://www.canada.ca/en/health-canada/services/drugs-health-products/covid19-industry/drugs-vaccines-treatments/vaccines.html>, the following list of COVID-19 vaccines are currently being authorized by Health Canada:

|      COVID-19 Vaccine       |           WHO Name            |      OWID Name       | Date of Approval |
| :-------------------------: | :---------------------------: | :------------------: | :--------------: |
|  Pfizer-BioNTech Comirnaty  | `Pfizer BioNTech - Comirnaty` |  `Pfizer/BioNTech`   |  **2020-12-09**  |
|      Moderna Spikevax       |     `Moderna - Spikevax`      |      `Moderna`       |  **2020-12-23**  |
|    AstraZeneca Vaxzevria    |   `AstraZeneca - Vaxzevria`   | `Oxford/AstraZeneca` |  **2021-02-26**  |
| Janssen (Johnson & Johnson) |   `Janssen - Ad26.COV 2-S`    |  `Johnson&Johnson`   |  **2021-03-05**  |
|      Novavax Nuvaxovid      |      `Novavax-NUVAXOVID`      |    **`Novavax`**     |  **2022-02-17**  |
|      Medicago Covifenz      |               /               |  (**`Medicago`**?)   |  **2022-02-24**  |

### Why do you think the data is inaccurate or missing?

1. Since all the above COVID-19 vaccines are currently being authorized under the *Food and Drugs Regulations*, which equals full authorization in many other countries, it is necessary to add all of them to the list of COVID-19 vaccines used in Canada.
2. Medicago Covifenz COVID-19 vaccine doesn't have the corresponding WHO name up to now, and it could be called `Medicago` like other vaccines.

lucasrodes:
Hi @aywi,
Thanks for your help once again.

Currently, both Novavax and Medicago are not listed in our source (https://covid19tracker.ca/vaccinationtracker.html).

Perhaps @noahlittle has some insights?

aywi:
> Hi @aywi, Thanks for your help once again.
> 
> Currently, both Novavax and Medicago are not listed in our source (https://covid19tracker.ca/vaccinationtracker.html).
> 
> Perhaps @noahlittle has some insights?

[...]

PR #2619, opened 2022-04-27
feat(vax,can): COVID-19 Vaccines Currently Authorized by Health Canada

aywi:
Solve #2618.


aywi:
Close for further investigation.

PR #2620, opened 2022-04-29
feat(vax,chn): Reduce running time and still ensure data validity

aywi:
See the title.


scripts/src/cowidev/vax/incremental/china.py
```diff
@@ -72,11 +71,14 @@ def read_complete(self):
...
             Wait(driver, self.timeout).until(EC.presence_of_element_located((By.CLASS_NAME, "zxxx_list")))
-            time.sleep(random.randint(2, 3))
             driver.execute_script("window.stop();")
             links = self._get_links_complete(driver)
             for link in links[: self.num_links_complete]:
-                record = self._parse_data_complete(driver, link)
+                try:
+                    record = self._parse_data_complete(driver, link)
+                except:
+                    print("Failed to parse:", link)
```
aywi, line 80: I don't know whether "print" is a right statement to use in this project when I want to output an acceptable error.

aywi: @lucasrodes Thanks! Since the only possible exception in the subroutine is TimeoutException (in two different locations), I will cancel the try-except and let them happen in the subroutine. See #2621.

PR #2621, opened 2022-04-29
feat(vax,chn): Don't skip exceptions, and the manual entry for China update on 2022-04-28

aywi:
See the title. And the manual entry for China update on 2022-04-28, see https://github.com/owid/covid-19-data/issues/230#issuecomment-1113466700.


lucasrodes:
LGTM

PR #2622, opened 2022-04-30
fix(jhu): Hide China death data correction on 2020-04-17

aywi:
Source: <http://www.nhc.gov.cn/xcs/yqtb/202004/9d15772389c64d478713e710a756b883.shtml>.


edomt:
Good point, thank you!

PR #2624, opened 2022-05-01
feat(vax,can): Add Novavax and correct vaccine timeline & feat(vax): Add Medicago

aywi:
See #2618.


aywi:
Close this one in favor of #2641.

Issue #2627, opened 2022-05-02
data(vax,chn): Problematic data on 2022-04-30 and 2022-05-02


aywi:
### Country

China

### Domain

Vaccinations

### Which data is inaccurate or missing?

1. The data on 2022-04-30 is missing because the list page (current source) <http://www.nhc.gov.cn/xcs/yqjzqk/list_gzbd.shtml> does not include the corresponding data. But I can find the [data on 2022-04-30](http://www.nhc.gov.cn/jkj/s7915/202205/a812fd406ee74cc6b35780d9abac0a74.shtml) from another list page <http://www.nhc.gov.cn/jkj/pzhgli/new_list.shtml>. It is also the list page holding actual HTML pages for the data since 2021-07-28. We can safely switch to this new source.

2. The page refreshing problem still happens on the server-side, therefore the source_url for the data on 2022-05-02 is incorrect.

### Why do you think the data is inaccurate or missing?

1. Switch to the new source <http://www.nhc.gov.cn/jkj/pzhgli/new_list.shtml> for the read() in the current script.
2. Use other methods to avoid potential page loading/refreshing problems instead of using time.sleep().

I have submitted #2629. I will be monitoring the new source for one more week:
- [x] [2022-05-04](https://github.com/owid/covid-19-data/blob/f46e965651e666d6bfbffea52343478685917cb7/scripts/output/vaccinations/status/status-vax-get.csv#L58) (28.51s)
- [x] [2022-05-05](https://github.com/owid/covid-19-data/blob/071c105cd8e948a744a796323b97886ed52a1500/scripts/output/vaccinations/status/status-vax-get.csv#L57) (96.16s)
- [x] [2022-05-06](https://github.com/owid/covid-19-data/blob/772fb63f05810a0d6d5024c9f6cee2bc9c47a40e/scripts/output/vaccinations/status/status-vax-get.csv#L57) (58.0s)
- [x] [2022-05-07](https://github.com/owid/covid-19-data/blob/bbbaf101b7f7cc58d32061d3f641fb3ee2b0b920/scripts/output/vaccinations/status/status-vax-get.csv#L57) (48.0s)
- [x] [2022-05-08](https://github.com/owid/covid-19-data/blob/4504eadea5fdd89dc982ea75723d575c6cd3ce9b/scripts/output/vaccinations/status/status-vax-get.csv#L57) (32.04s)
[...]

PR #2629, opened 2022-05-03
feat(vax,chn): Switch to the new source and avoid the page refreshing problem

aywi:
Also fix the data on 2022-04-30 and 2022-05-02. See #2627.


edomt:
Thank you @aywi !

Issue #2630, opened 2022-05-03
data(vax,ecdc): Additional EU/EEA Countries That Have Administered Novavax Nuvaxovid COVID-19 Vaccine


aywi:
### Country

ECDC

### Domain

Vaccinations

### Which data is inaccurate or missing?

According to the [COVID-19 Vaccine Tracker of ECDC][COVID-19-Vaccine-Tracker-of-ECDC], the following list of EU/EEA countries have administered Novavax Nuvaxovid COVID-19 vaccine but haven't been included in the OWID vaccination dataset:

|  Country   | Vaccine Timeline  |           Administration            |
| :--------: | :---------------: | :---------------------------------: |
|  Belgium   |   Batch (Fixed)   | [2022-01-19][2022-01-19-Belgium][1] |
|  Croatia   |    Incremental    |  [2022-03-11][2022-03-11-Croatia]   |
|   Cyprus   |    Incremental    |   [2022-03-11][2022-03-11-Cyprus]   |
|  Estonia   |   Batch (Fixed)   |  [2022-03-04][2022-03-04-Estonia]   |
|   France   | Batch (Automated) | [2022-02-26][2022-02-26-France][1]  |
|   Greece   |   Batch (Fixed)   |        [2022-03-02][WHO][2]         |
|  Ireland   |   Batch (Fixed)   | [2022-02-04][2022-02-04-Ireland][3] |
|   Latvia   | Batch (Automated) | [2022-02-28][2022-02-28-Latvia][1]  |
| Luxembourg |   Batch (Fixed)   | [2022-03-04][2022-03-04-Luxembourg] |
|   Poland   |    Incremental    |   [2022-03-04][2022-03-04-Poland]   |
|   Sweden   |   Batch (Fixed)   |   [2022-03-11][2022-03-11-Sweden]   |

- [1] From the individual country's source.
- [2] The date is from WHO while is unknown from [ECDC open data][ECDC-open-data].
- [3] Has ignored 2 abnormal doses which are way before the [approval of Novavax in EU (on 2021-12-20)][approval-of-Novavax-in-EU] (possibly from clinical trials).

There are some additional problems for France and Latvia since their vaccine timelines are automatically generated:
1. For France, the [current source][2022-02-26-France] does include Novavax, but this isn't reflected in the current script;
[...]

PR #2634, opened 2022-05-04
fix(vax,fra): Add Pfizer/BioNTech pediatric version and Novavax

aywi:
Update according to <https://www.data.gouv.fr/fr/datasets/donnees-relatives-aux-personnes-vaccinees-contre-la-covid-19-1/>.

This will affect the data since 2021-12-14. See also #2630.


edomt:
Thank you @aywi, very useful as always!

The discrepancy you spotted is due to the way that the French government calculates these metrics. Their current policy is basically that one infection = one dose, and they're calculating the metrics this way, which makes for slightly different results compared to us (because we only count actual vaccinations, not infections).

PR #2641, opened 2022-05-08
feat(vax,can): Add age group and manufacturer data from Public Health Agency of Canada

aywi:
See #2618.


lucasrodes:
Looks good to me!

We are combining your data with data from https://health-infobase.canada.ca, @noahlittle, to obtain the dates when each vaccine started to be used. We observe that Novavax and J&J appear there but not on your data.

Based on [your comment](https://github.com/owid/covid-19-data/issues/2618#issuecomment-1111196446), do you think this may be because your data is based on provincial data while infobase's is at national level?

aywi:
@lucasrodes Actually from [covid19tracker.ca](https://covid19tracker.ca/vaccinationtracker.html) I only saw vaccines **distributed** (and not administered) by vaccine types and the last update is 2022-02-03. I guess this is because not all provincial sources provide this kind of data (at least [British Columbia](http://www.bccdc.ca/health-info/diseases-conditions/covid-19/data) is not).
Their data group data are also from PHAC. When I checked [the API](https://api.covid19tracker.ca/docs/1.0/vaccine-age-groups), I found the last update is 2022-01-30.

aywi:
> @lucasrodes Actually from [covid19tracker.ca](https://covid19tracker.ca/vaccinationtracker.html) I only saw vaccines **distributed** (and not administered) by vaccine type and the last update is 2022-02-03. I guess this is because not all provincial sources provide doses administered by vaccine type data now (at least [British Columbia](http://www.bccdc.ca/health-info/diseases-conditions/covid-19/data) is not). Their data group data are also from PHAC. When I checked [the API](https://api.covid19tracker.ca/docs/1.0/vaccine-age-groups) (with different subgroups), I found the last update is 2022-01-30.

This issue mentioned this: https://github.com/andrewthong/covid19tracker-api/issues/67

lucasrodes:
LGTM! I will merge as I think having vaccinated by age and by the manufacturer is a huge plus!

I have added some checks to detect possible column/field changes!

Thanks for the exhaustive research.

PR #2645, opened 2022-05-12
fix(vax,fra): Fix source_url of manufacturer data

aywi:
https://github.com/owid/covid-19-data/blob/d0ccd1e1de71d81ffb59ba28bac05e55e846f5a3/scripts/src/cowidev/vax/batch/france.py#L99

It should be `self.source_url_ref`, not `self.source_url`.


lucasrodes (approved):
Thanks, @aywi!

PR #2648, opened 2022-05-13
fix(vax,chn): Fix regex for data on 2022-05-12

aywi:
This time NHC China breaks the summary sentence it always used before. The new version is still compatible with all previous data. Other changes are just refactoring.


lucasrodes:
LGTM, thanks!

PR #2650, opened 2022-05-14
fix(vax,can): Add new columns for age group and manufacturer data

aywi:
- Age group data: PHAC started recording `people_with_2nd_booster`. Since OWID hasn't used this metric yet, I will leave it alone.
- Manufacturer data: PHAC starts to distinguish between dose 4 and dose 5+.


lucasrodes:
Thanks!

PR #2654, opened 2022-05-18
fix(vax,cze): Remove unapproved vaccines

aywi:
Sinopharm/Beijing and Sinovac have never been approved in Czechia, and their very few doses are possibly attributed to the people received in other countries. Same as Latvia (#2637).


lucasrodes:
Thanks!

PR #2655, opened 2022-05-18
enhance(vax,can): Better logic to deal with make_monotonic exceptions

aywi:
The new logic will allow fetching data for the last two dates if no make_monotonic exception occurs.


lucasrodes:
LGTM - thanks!

PR #2660, opened 2022-05-23
fix(vax,chn): ❗ Redesign regex & merge pipeline

aywi:
The regex breaks again for the total_boosters on 2022-05-22 (you will see a significant drop tomorrow). It is now necessary to redesign the regex and the merge pipeline:
1. The new regex only focuses on keywords and tolerates the potential changes in the Chinese language as many as possible: https://github.com/owid/covid-19-data/blob/9a346b8e95de5048b3ba1e7af102cfc9b419ef91/scripts/src/cowidev/vax/incremental/china.py#L22-L32
2. The new merge pipeline avoids saving the inaccurate total_vaccinations or clearing previously saved metrics or saving abnormally decreased metrics: https://github.com/owid/covid-19-data/blob/9a346b8e95de5048b3ba1e7af102cfc9b419ef91/scripts/src/cowidev/vax/incremental/china.py#L135-L156
3. The previous source goes back to normal. Since it is faster, it is better to switch back: https://github.com/owid/covid-19-data/blob/9a346b8e95de5048b3ba1e7af102cfc9b419ef91/scripts/src/cowidev/vax/incremental/china.py#L15

### DIFF
--- a/scripts/src/cowidev/vax/incremental/china.py
+++ b/scripts/src/cowidev/vax/incremental/china.py
@@ -12,19 +12,23 @@ from cowidev.vax.utils.base import CountryVaxBase
 
 class China(CountryVaxBase):
     location: str = "China"
-    source_url: str = "http://www.nhc.gov.cn/jkj/pzhgli/new_list.shtml"
+    source_url: str = "http://www.nhc.gov.cn/xcs/yqjzqk/list_gzbd.shtml"
     source_url_complete: str = "http://www.nhc.gov.cn/xcs/s2906/new_list.shtml"
     regex: dict = {
         "title": "新冠病毒疫苗接种情况",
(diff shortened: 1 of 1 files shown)


aywi:
@edomt @lucasrodes Please take a look since the bug will bring a significant drop to the world total_boosters tomorrow.

lucasrodes:
Thanks for working on this, @aywi.

FYI, we have [some checks](https://github.com/owid/covid-19-data/blob/bf58d13b2cc078fe884751f93ce288b551e0c03a/scripts/src/cowidev/vax/utils/checks.py#L171-L182) to avoid drops in metrics to be propagated to our public files, so this would not have been shown on our live charts or public datasets. This is executed in `cowid vax process` step.

PR #2662, opened 2022-05-24
enhance(vax,jpn): Batch version with age group and manufacturer data

aywi:
The previous batch version [`japan.py`](https://github.com/owid/covid-19-data/blob/a7dc88c2181fcb4bdcd370677b183f019a0f233d/scripts/scripts/vaccinations/src/vax/batch/japan.py) was initially created by @fqj1994. I don't know the last reason to abandon it (maybe the inclusion of workplace data?), but since Japan's data are usually backdated as mentioned in #1608 and the data format has been stable since [2022-03-02](https://web.archive.org/web/20220302213952/https://www.kantei.go.jp/jp/headline/kansensho/vaccine.html), it is better to switch to batch version now.

The new batch version now has:
- Much more accurate daily metrics.
- Age group data. Currently only has age 5-11, 12-64, and 65+ groups because only these three age groups are provided with daily metrics (but is better than none).
- Manufacturer data. The last update was 2021-08-04: https://github.com/owid/covid-19-data/blob/253fd1e711b33770a2c88b6dc3beb0f07cdb095a/scripts/output/vaccinations/by_manufacturer/Japan.csv#L225-L226
- Sanity checks for new sheets (unlikely), new columns (like dose 4) and new vaccines (like Novavax). It is unable to detect new age groups now, but this won't break the main data.


fqj1994:
Basically, the reason why batch is dropped is the introduction of workplace vaccination.
Since the overlap/duplicate is not backdated, there is no accurate way to calculate vaccination doses in any past days.

aywi:
> Basically, the reason why batch is dropped is the introduction of workplace vaccination. Since the overlap/duplicate is not backdated, there is no accurate way to calculate vaccination doses in any past days.

Yes, the `overlap` data are not backdated. But the latest (2022-05-24) `workplace - overlap` dose 1 & dose 2 cumsum = 19,397,565 - 19,273,729 = 123,836 doses, which is even much smaller than the one day increment 462,291 doses. That means most `workplace` doses (>99%) have already been included in `general` daily doses and the effect of remaining uncounted doses is very small.

Even for these 123,836 uncounted `workplace` doses, we can do a fair assumption that they are evenly distributed in 19,397,565 overall `workplace` doses. Because the day 1 (2021-08-01) of `workplace` doses have already been 7,723,380, which is much larger than 123,836 and also means `workplace` doses have been gradually included in `general` daily doses starting from day 1. The corresponding code is here: https://github.com/owid/covid-19-data/blob/74c30416296c633d882fb6ef1318c85c0c864d83/scripts/src/cowidev/vax/batch/japan.py#L105-L110

lucasrodes:
Thank you very much for your help. LGTM.

PS: I'll revisit age data on our front-end, as it may not all the age groups may be displayed for Japan (if they are unique to Japan, I need to 'activate' them)

aywi:
@lucasrodes Currently, the 12-64 and 65+ age groups are unique to Japan.
Also, I found that the 0-4 age group is not displayed for Canada. Is it disabled for all countries right now? (some EU countries also have)

lucasrodes:
Yes, it is disabled due to technical reasons. We have a limitation on the number of different age groups that we can have on our chart (considering all countries).

PR #2663, opened 2022-05-26
fix(vax,jpn): Add Novavax & 2nd booster

aywi:
Update on 2022-05-26:
- Novavax has started being recorded, although the total_vaccinations is 0 up to now.
- 2nd booster added.

One thought:
Since many countries start recording 2nd or more boosters, and the people_with_booster has been used in age group data, maybe the people_with_booster is also needed in main data to distinguish against total_boosters. Because the total_boosters is a dose-based metric like total_vaccinations, but the people_with_booster is a coverage-based metric like people_vaccinated & people_fully_vaccinated. Chile already has the total_boosters_per_hundred > 100, which may make many people think the coverage is also near 100%.


lucasrodes:
Thanks for your contribution.

On your thought, we are not planning on adding new metrics to the vaccination dataset. It is very costly, and for now, this is not a priority. However, we may revisit this in the future. Thanks for your suggestion.

PR #2664, opened 2022-05-27
fix(vax,jpn): Remove age groups for 2nd booster

aywi:
Since the 2nd booster is not useful for age group data and the data format of the 2nd booster is still new & unstable in recent days.

Also: The 12-64 and 65+ age groups are not yet displayed for Japan.


lucasrodes:
LGTM - thanks!

All group ages for Japan should be live on our site.

aywi:
@lucasrodes Great! I guess there is an annotation missing:

Japan_people_with_booster_age_groups

lucasrodes:
Thanks for reporting! Fixed

PR #2665, opened 2022-05-27
fix(vax,can): Use changes instead of totals

aywi:
If there are some provinces that haven't been reported in a day (which often happens recently), the totals for the latest day fetched from the API are incomplete (only provinces with updates are included). Use changes can get the exact latest metrics as the https://covid19tracker.ca/vaccinationtracker.html.

### DIFF
--- a/scripts/src/cowidev/vax/batch/canada.py
+++ b/scripts/src/cowidev/vax/batch/canada.py
@@ -84,10 +84,10 @@ class Canada(CountryVaxBase):
                 "total_vaccines_distributed",
             ],
         )
-        return df[["date", "total_vaccinations", "total_vaccinated", "total_boosters_1", "total_boosters_2"]]
+        return df[["date", "change_vaccinations", "change_vaccinated", "change_boosters_1", "change_boosters_2"]]
 
     def read_age(self) -> pd.DataFrame:
-        df = read_csv_from_url(self.source_url_a, verify=False)
+        df = read_csv_from_url(self.source_url_a)
         check_known_columns(
             df,
             [
@@ -112,7 +112,7 @@ class Canada(CountryVaxBase):
         return df
 
     def read_manufacturer(self) -> pd.DataFrame:
-        df = read_csv_from_url(self.source_url_m, verify=False)
+        df = read_csv_from_url(self.source_url_m)
         check_known_columns(
             df,
             [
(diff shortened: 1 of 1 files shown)


lucasrodes:
Hi @aywi, thanks for your help.

I can see that using `change_` metrics can be tricky sometimes, especially when there are day gaps. 

Could you provide an example of how this 'incomplete' totals are shown in our data?

aywi:
> Hi @aywi, thanks for your help.
> 
> I can see that using `change_` metrics can be tricky sometimes, especially when there are day gaps. 
> 
> Could you provide an example of how this 'incomplete' totals are shown in our data?

Actually `change_` metrics are not tricky because I can reconstruct the same whole series with cumsum, except for the latest day. Day gaps don't matter because the changes only compare for dates that are in records. But it does require you to do the filtering after you got the cumcums. That's what I do in the pipe_get_totals function.

Only difference is in the latest day. For example, the data for the latest 2022-05-25, when I checked 5 hours ago, are only include provinces that have been reported for the `total_` metrics. You can check their website for the list of provinces. For the `change_` metrics, it looks like they used different logic for the latest day because metrics only include changes already reported, not current total record - previous total record (which works for other days). This logic actually helps because right now there are only limited provinces to report each day. That's the same logic they use to display latest metrics on their website.

Previously, because the incomplete data are usually too small, they will be filtered by make_monotonic. But there are some extremed cases when only some lower population provinces haven't been reported.

aywi:
@lucasrodes It happens again. You cannot use `total_` metrics to get the updates since 2022-05-30:

```python
data = request_json(Canada().source_url)
df = pd.DataFrame.from_records(data["data"])
df["change_vaccinations_cumsum"] = df.fillna(0).change_vaccinations.cumsum()
with pd.option_context("display.width", 100):
    print(
        df[df.total_vaccinations > 0][
            ["date", "change_vaccinations", "change_vaccinations_cumsum", "total_vaccinations"]
        ]
    )
```
```
           date  change_vaccinations  change_vaccinations_cumsum  total_vaccinations
324  2020-12-14                    5                           5                   5
325  2020-12-15                  721                         726                 726
326  2020-12-16                 2301                        3027                3027
327  2020-12-17                 4255                        7282                7282
328  2020-12-18                 4015                       11297               11297
..          ...                  ...                         ...                 ...
853  2022-05-27                32079                    85391529            85391529
854  2022-05-28                29441                    85420970            85420970
855  2022-05-29                20892                    85441862            85441862
856  2022-05-30                 8567                    85450429            53583333
857  2022-05-31                19508                    85469937            85469937

[534 rows x 4 columns]
```

In summary, `change_` metrics work in more cases than `total_` metrics.

lucasrodes:
LGTM, thanks for your work @aywi

PR #2682, opened 2022-06-09
fix(vax,jpn): Urgent fix for the drop of most rows today

aywi:
https://github.com/owid/covid-19-data/commit/2335bb28c0a4fa919b548704b7b79a0c71962773#diff-41d77541de49953c350a457ac29f85165de28e3b3443ce82c770a9f16024bc0d

The reason is the format change of NaN strings in one of the Excel sheets (from "―" to "ー"), which causes the drop of the doses 1 & 2 columns for most rows. The current way of handling NaN strings cannot deal with this change. Will replace it with a much more robust method: https://github.com/owid/covid-19-data/blob/f99d3f74a837db10b12820c76816d1a115de7e73/scripts/src/cowidev/vax/batch/japan.py#L145

Others are just refactoring.


aywi:
@edomt @lucasrodes https://github.com/owid/covid-19-data/commit/a5f2230a586554fc62928b3843b21b91db6b68be It is actually the bug of original code (only consider fixed-form NaN strings). It is now fixed.

lucasrodes:
Thank you very much, @aywi!

PR #2690, opened 2022-06-20
fix(vax,chn): Update waiting conditions

aywi:
It is weird that the complete data on 2022-06-16 cannot be fetched on the server-side, but nothing wrong locally. A potential reason is that the waiting conditions are not accurate enough, so the main content may not be fully loaded (when the connection speed is unstable). Here are the waiting conditions that ensure the main content is fully loaded while other unnecessary (slow) web elements are still ignored.


lucasrodes:
Let's try these conditions from now on. They worked fine for me.

Thanks @aywi !

PR #2703, opened 2022-07-23
fix(vax,can): Add Medicago

aywi:
It has been administered in Canada (doesn't have WHO name yet).


lucasrodes:
Thanks, @aywi!

@noahlittle are you aware of this new vaccine being administered in Canada? It is not showing in https://covid19tracker.ca/vaccinationtracker.html

lucasrodes (approved):

PR #2707, opened 2022-08-03
Fix/server errors

lucasrodes:
Currently there are some of the pipeline scripts that are not working in the server (but on our local machines)

**Vaccinations:**

- [x] Japan: Something with the URL is wrong
- [ ] ECDC: Timeout
- [ ] Russia: Returns 502 in the server

**Hospitalizations**

- [ ] ECDC not working, which makes the whole pipeline break


lucasrodes:
Merging only with fix for Japan.

Some findings:
- For ECDC, the connection can't be established from our Digital Ocean bucket (London-based) but it works if we use a NY-based one. Command: `wget http://opendata.ecdc.europa.eu/covid19/vaccine_tracker/csv/data.csv`
- For Russia, it is unclear what is happening

PR #2789, opened 2023-02-21
Megafile(cases,deaths): Update tools to use data from WHO 

lucasrodes:
Currently, we use the output from `cowid jhu generate` for the cases and deaths metrics processing in the `megafile` step (`cowid megafile`).

We must replace this with the output from `cowid casedeath generate`.


lucasrodes:
`cowid megafile` will now generate outputs based on WHO's data. For the time being, nothing changes on the main datasets, however new ones have been added:

- `public/data/owid-covid-data-new.csv`: Same as `public/data/owid-covid-data.csv` but using WHO's data.
- `public/data/internal_new`: Same as `public/data/internal` but using WHO's data. Only affected files are added.

lucasrodes:
This PR also adds some minor fixes to the pipeline. E.g., country names.

PR #2805, opened 2023-06-14
:hammer: get excess mortality from Economist from ETL catalog

Marigold:
Fetch excess mortality estimates from The Economist from ETL instead of updating it manually.

### DIFF
--- a/scripts/src/cowidev/xm/etl.py
+++ b/scripts/src/cowidev/xm/etl.py
@@ -113,7 +113,6 @@ class XMortalityEconomistETL:
         df_current = pd.read_csv(output_path)
         if not df.equals(df_current):
             # Export data
-            __import__('ipdb').set_trace()
             df.to_csv(output_path, index=False)
 
     def run(self):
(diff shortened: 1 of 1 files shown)


scripts/src/cowidev/xm/etl.py
```diff
@@ -102,6 +99,32 @@ def run(self):
...
+        return pd.DataFrame(t), date_accessed
+
+    def transform(self, df: pd.DataFrame) -> pd.DataFrame:
+        return df.sort_values(["country", "date"])
+
+    def load(self, df: pd.DataFrame, output_path: str, date_accessed: str) -> None:
+        df_current = pd.read_csv(output_path)
+        if not df.equals(df_current):
+            # Export data
+            __import__('ipdb').set_trace()
```
lucasrodes, line 116: what is this magic?

lucasrodes (approved):
LGTM! Thanks for doing this, @Marigold

scripts/src/cowidev/xm/etl.py
Marigold: That's a forgotten breakpoint... which is why we're doing reviews!

PR #2807, opened 2023-07-19
:hammer: remove grapher DB code

Marigold:
Could be merged after we get it to ETL with https://github.com/owid/etl/pull/1361. The PR removes all mentions of grapher-db and its code.

Some changes are from automatic black / isort formatting (using proper line length from black settings). I could also create a PR that blacks everything first if you don't want it in the same PR.

## TODO
* [x] test that everything works
* [x] update ops scripts to stop calling grapher code


Marigold:
@lucasrodes all COVID datasets now consume .csv files generated by this repo and are processed in ETL. This PR only removes code that handles upserts to MySQL. What's the right way to test this, please?

lucasrodes:
Hi @Marigold,
Thanks for doing this!

I think that the most straightforward way to test this is to ssh into the covid `lxc` machine, run the different code blocks from [autoupdate.sh](https://github.com/owid/covid-19-data/blob/master/scripts/scripts/autoupdate.sh) and confirm that it runs without any errors.

Marigold:
@lucasrodes I tested it ad-hoc and the entire autoupdate.sh seems to be working. Could you take a quick look and approve if you think it's ok?

lucasrodes (approved):
Once again, thanks for doing this, @Marigold.

LGTM, let's drop the bomb

